PR #4, opened 2017-06-05
Use no user when the users section is missing in kubeconfig

pokoli:
Fixes kubernetes-incubator/client-python#236

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -130,7 +130,8 @@ class KubeConfigLoader(object):
             context_name = self._config['current-context']
         self._current_context = self._config['contexts'].get_with_name(
             context_name)
-        if self._current_context['context'].safe_get('user'):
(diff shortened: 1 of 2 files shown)


mbohlool:
Looks good. Thanks.

config/kube_config.py
```diff
@@ -130,7 +130,8 @@ def set_active_context(self, context_name=None):
             context_name = self._config['current-context']
         self._current_context = self._config['contexts'].get_with_name(
             context_name)
-        if self._current_context['context'].safe_get('user'):
+        if (self._current_context['context'].safe_get('user')
+                and self._config.safe_get('users')):
```
mbohlool: Actually we can protect this more by also checking if `users` section has the specified user or not.

pokoli: Created #5 in order to fix your comment. Thanks

PR #5, opened 2017-06-06
Use no user when the especified user is not found in the users section

pokoli:


config/kube_config.py
```diff
@@ -132,8 +132,12 @@ def set_active_context(self, context_name=None):
             context_name)
         if (self._current_context['context'].safe_get('user')
                 and self._config.safe_get('users')):
-            self._user = self._config['users'].get_with_name(
-                self._current_context['context']['user'])['user']
+            users = dict((x['name'], x) for x in self._config['users'])
```
mbohlool: I don't think you need to construct this dict. You can just add another condition to the if statement, something like this `self._config['users'].get_with_name(self._current_context['context'].safe_get('user')` or you can store this into a variable and reuse it if we don't want get_with_name to be called twice.

pokoli: This won't prevent the case where the user name is not in the user list as the get_with_name raises an error [if name not found](https://github.com/kubernetes-client/python-base/blob/master/config/kube_config.py#L273). And that's why I build the dict.

Another option is to add the safe, parameter to the `get_with_name` function, which will make it return None when not found, which will be disabled by default and only enabled here. 

What do you think?

mbohlool: safe parameter for get_with_name is a good idea.

pokoli: Added.

PR #6, opened 2017-06-06
Add travis and tox

pokoli:
Closes #3

PR #9, opened 2017-06-08
Add tox and travis support

### LINKED ISSUE #3: Add tox and travis
We should add travis and tox to ensure test pass on all the suported platforms (py27, py34, py35)

I'm wondering if we should also add coverage and pep8 checks

I can work on it

mbohlool:
fixes #3

based on #6 by @pokoli


tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
+skipsdist = True
+envlist = py
+
+[testenv]
```
pokoli: What about coverage?

mbohlool:
@pokoli this is ready now. Thanks for the review.

api_client.py
```diff
@@ -1,45 +1,38 @@
-# coding: utf-8
-
-"""
-Copyright 2016 SmartBear Software
```
pokoli: Are we sure we can remove this copy right?

run_tox.sh
```diff
@@ -0,0 +1,57 @@
...
+SCRIPT_ROOT=`pwd`
+popd > /dev/null
+
+cd "${TMP_DIR}"
+git clone https://github.com/kubernetes-incubator/client-python.git
+cd client-python
+git config user.email "kubernetes-client@k8s.com"
+git config user.name "kubenetes client"
+git rm -rf kubernetes/base
+git commit -m "DO NOT MERGE, removing submodule for testing only"
```
pokoli: I'm wondering why is required to commit, can we avoid it?

pokoli:
@mbohlool Looks good to me. I added two minor comments, but for me it can be merged. Good work.

run_tox.sh
mbohlool: Update-pep8 check will fail if there is an uncommited change.

api_client.py
mbohlool: Good catch. I prefer not to remove that. Will fix.

run_tox.sh
pokoli: Acknowledged. I was not aware of this part! Great.

tox.ini
mbohlool: All environment from .travis file will be passed to the main repo tox, coverage is running too (codecov is already commenting on the PR).

pokoli:
LGTM

PR #8, opened 2017-06-08
added configuration 'http_proxy' to allow the usage of a proxy

mbohlool:
porting https://github.com/kubernetes-incubator/client-python/pull/206 to python-base repo.

PR #15, opened 2017-06-13
Add http_proxy configuration with unittest

pokoli:
This PR extends #8 by adding some unit test for the PoolManager and the proxy manager.

### DIFF
--- a/rest_test.py
+++ b/rest_test.py
@@ -16,7 +16,7 @@ import urllib3
 
 from mock import patch
 
-from kubernetes.client import Configuration
+from kubernetes.client import ConfigurationObject
 from kubernetes.client.rest import RESTClientObject
 
 
@@ -25,12 +25,12 @@ class RestTest(unittest.TestCase):
     def test_poolmanager(self):
         'Test that a poolmanager is created for rest client'
         with patch.object(urllib3, 'PoolManager') as pool:
-            RESTClientObject(config=Configuration())
+            RESTClientObject(config=ConfigurationObject())
             pool.assert_called_once()
 
     def test_proxy(self):
         'Test that proxy is created when the config especifies it'
-        config = Configuration()
+        config = ConfigurationObject()
         config.http_proxy_url = 'http://proxy.example.com'
 
         with patch.object(urllib3, 'proxy_from_url') as proxy:
--- a/rest_test.py
+++ b/rest_test.py
@@ -1,0 +1,42 @@
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import unittest
+import urllib3
+
+from mock import patch
+
+from kubernetes.client import Configuration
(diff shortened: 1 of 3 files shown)


mbohlool:
@pokoli Sorry that I missed this. Please ping me if I missed a PR like this ever in future. It should not take me more than a couple of days to do a code review. More than that is not usual :)

mbohlool:
Please resolve the conflicts and ping me after.

rest_test.py
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import unittest
+import urllib3
+
+from mock import patch
+
+from .configuration import configuration
```
mbohlool: Get this from kubernetes package like if you installed the package. CI tests expect that.

pokoli: Done, thanks for the hint.

pokoli:
Running on `env TOXENV=py27 ./run_tox.sh tox` works as expected, but this si failing on travis. 

@mbohlool Any idea?

mbohlool:
`from kubernetes.client import configuration`

`base` folder is not a proper python package.

pokoli:
Uou, so much days without kubernetes-client lost my memory :( @mbohlool  Thanks for the tips

Now only functional tests are failing, but I dont' think this is related to my change.

mbohlool:
Closing and reopening to see if it helps with ci failure.

rest_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_poolmanager(self):
+        'Test that a poolmanager is created for rest client'
+        with patch.object(urllib3, 'PoolManager') as pool:
+            RESTClientObject(config=configuration)
+            pool.assert_called_once()
+
+    def test_proxy(self):
+        'Test that proxy is created when the config especifies it'
+        configuration.http_proxy_url = 'http://proxy.example.com'
```
mbohlool: I think this is the problem, you are changing configuration singleton that has effect on other clients. You should create your own version of configuration:
```python
from kubernetes.client import Configuration
config=Configuration()
...
```

pokoli: That makes sense, but it does not fix the problem. 

Indeed, running py27-functional on my laptop passes.

mbohlool: The way e2e tests are setup here is not the best. They will pass if there is no cluster to run on. You need to make sure you have a minikube running on default port before run functional tests. We need to fix that but let see what is the problem of this PR.

mbohlool: So the problem is still Configuration(). It should be ConfigurationObject(). Configuration() is there for backward compatibility with generated client and still returns the singleton. using ConfigurationObject() should fix things.

rest_test.py
```diff
@@ -0,0 +1,42 @@
...
+from kubernetes.client import Configuration
+from kubernetes.client.rest import RESTClientObject
+
+
+class RestTest(unittest.TestCase):
+
+    def test_poolmanager(self):
+        'Test that a poolmanager is created for rest client'
+        with patch.object(urllib3, 'PoolManager') as pool:
+            RESTClientObject(config=Configuration())
```
mbohlool: do you mean to pass a new config or singleton config here? `Configuration()` is the singleton. `ConfigurationObject()` will be a new one.

rest_test.py
```diff
@@ -0,0 +1,42 @@
...
+            pool.assert_called_once()
+
+    def test_proxy(self):
+        'Test that proxy is created when the config especifies it'
+        config = Configuration()
```
mbohlool: `config = ConfigurationObject()`

mbohlool:
I pushed a change to use ConfigurationObject and merged this because I want it to be in the release.

pokoli:
@mbohlool thank you for your help and for finishing it!

PR #18, opened 2017-07-09
Allow setting maxsize for PoolManager

yuvipanda:
Fixes https://github.com/kubernetes-incubator/client-python/issues/290.


yuvipanda:
I'm guessing I've to manually bump the submodule ref in client-python after this gets merged?

configuration.py
```diff
@@ -85,6 +83,12 @@ def __init__(self):
         # Set this to True/False to enable/disable SSL hostname verification.
         self.assert_hostname = None
 
+        # urllib3 connection pool's maximum number of connections saved
+        # per pool. Increasing this is useful for cases when you are
+        # making a lot of possibly parallel requests to the same host,
+        # which is often the case here.
+        self.connection_pool_maxsize = 1
```
mbohlool: This will change default value to 1. I don't know what default value is, but it is safer to set this to None and add an if statement in the other file to set it if not none.

yuvipanda: Per http://urllib3.readthedocs.io/en/latest/advanced-usage.html the default value is 1 but I agree we should use None and pass through. Shall amend.

yuvipanda: Done

PR #21, opened 2017-07-20
Fixing get Google default credentials method

gabrielgbim:
This should fix the following error:
```
Traceback (most recent call last):
  File "watch-namespace.py", line 4, in <module>
    config.load_kube_config()
  File "egg/kubernetes/config/kube_config.py", line 318, in load_kube_config
  File "egg/kubernetes/config/kube_config.py", line 221, in load_and_set
  File "egg/kubernetes/config/kube_config.py", line 160, in _load_authentication
  File "egg/kubernetes/config/kube_config.py", line 176, in _load_gcp_token
  File "egg/kubernetes/config/kube_config.py", line 124, in <lambda>
NameError: global name 'GoogleCredentials' is not defined
```

PR #22, opened 2017-07-21
Add proper GCP config loader and refresher

mbohlool:
This is based on #21. I added expiration mechanism as well as persisting config back so we don't need to refresh the token every time. The way it persisted back should be the same as `kubectl`.
Tests will fail for this PR unless https://github.com/kubernetes-incubator/client-python/pull/302 is merged.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -14,17 +14,21 @@
 
 import atexit
 import base64
+import datetime
 import os
 import tempfile
 
+import google.auth
+import google.auth.transport.requests
 import urllib3
 import yaml
-from google.oauth2.credentials import Credentials
 
 from kubernetes.client import ApiClient, ConfigurationObject, configuration
 
 from .config_exception import ConfigException
+from .dateutil import UTC, format_rfc3339, parse_rfc3339
 
+EXPIRY_SKEW_PREVENTION_DELAY = datetime.timedelta(minutes=5)
 KUBE_CONFIG_DEFAULT_LOCATION = os.environ.get('KUBECONFIG', '~/.kube/config')
 _temp_files = {}
 
@@ -52,6 +56,11 @@ def _create_temp_file_with_content(content):
     with open(name, 'wb') as fd:
         fd.write(content.encode() if isinstance(content, str) else content)
     return name
+
+
+def _is_expired(expiry):
+    return ((parse_rfc3339(expiry) + EXPIRY_SKEW_PREVENTION_DELAY) <=
+            datetime.datetime.utcnow().replace(tzinfo=UTC))
 
 
 class FileOrData(object):
@@ -110,19 +119,26 @@ class KubeConfigLoader(object):
     def __init__(self, config_dict, active_context=None,
                  get_google_credentials=None,
                  client_configuration=configuration,
-                 config_base_path=""):
+                 config_base_path="",
+                 config_persister=None):
         self._config = ConfigNode('kube-config', config_dict)
         self._current_context = None
         self._user = None
         self._cluster = None
         self.set_active_context(active_context)
         self._config_base_path = config_base_path
+        self._config_persister = config_persister
+
+        def _refresh_credentials():
+            credentials, project_id = google.auth.default()
+            request = google.auth.transport.requests.Request()
+            credentials.refresh(request)
+            return credentials
+
         if get_google_credentials:
             self._get_google_credentials = get_google_credentials
         else:
-            self._get_google_credentials = lambda: (
-                GoogleCredentials.get_application_default()
-                .get_access_token().access_token)
+            self._get_google_credentials = _refresh_credentials
         self._client_configuration = client_configuration
 
     def set_active_context(self, context_name=None):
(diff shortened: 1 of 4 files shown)


gabrielgbim:
Thanks!

mbohlool:
@gabrielgbim @pokoli This is a release blocker. Can you review it please.

config/kube_config.py
```diff
@@ -54,6 +59,17 @@ def _create_temp_file_with_content(content):
     return name
 
 
+def _is_expired(expiry):
+    tf = tf_from_timestamp(expiry)
+    n = time.time()
```
pokoli: I don't think you need to store n as a variable.

config/kube_config.py
```diff
@@ -308,21 +353,32 @@ def load_kube_config(config_file=None, context=None,
         from config file will be used.
     :param client_configuration: The kubernetes.client.ConfigurationObject to
         set configs to.
+    :param persist_config: If True and config changed (e.g. GCP token refresh)
```
pokoli: Maybe better: "IF True config file will be updated when changed (e.g GCP token refresh)

config/kube_config.py
```diff
@@ -308,21 +353,32 @@ def load_kube_config(config_file=None, context=None,
...
+        the provided config file will be updated.
     """
 
     if config_file is None:
         config_file = os.path.expanduser(KUBE_CONFIG_DEFAULT_LOCATION)
 
+    config_persister = None
+    if persist_config:
+        config_persister = lambda config_map, config_file=config_file: (
```
pokoli: Why is required a lambda function here?

config/kube_config_test.py
```diff
@@ -464,24 +490,38 @@ def test_load_user_token(self):
...
             host=TEST_HOST,
-            token=BEARER_TOKEN_FORMAT % TEST_ANOTHER_DATA_BASE64)
+            token=BEARER_TOKEN_FORMAT % TEST_DATA_BASE64)
         actual = FakeConfig()
         KubeConfigLoader(
             config_dict=self.TEST_KUBE_CONFIG,
             active_context="gcp",
             client_configuration=actual,
-            get_google_credentials=lambda: TEST_ANOTHER_DATA_BASE64) \
+            get_google_credentials=lambda: "SHOULD NOT BE CALLED") \
```
pokoli: maybe you can put a raise to ensure it's not called

config/rfc3339.MD
```diff
@@ -0,0 +1 @@
+The (rfc3339.py)[rfc3339.py] file is copied from [this site](http://home.blarg.net/~steveha/pyfeed.html) because PyFeed is not available in PyPi.
```
pokoli: What about https://pypi.python.org/pypi/rfc3339/5.2 (source code in https://bitbucket.org/henry/rfc3339/src)

mbohlool: looks like they only do one side of conversion? https://bitbucket.org/henry/rfc3339/issues/3/datetime-to-rfc-3339

I see a parse method but it is not exported.

pokoli: Then probably it's worth to contribute one to the library :)

config/kube_config.py
mbohlool: Fixed.

config/kube_config.py
mbohlool: Fixed!

config/kube_config_test.py
mbohlool: to pass the `config_file` to `_save_kube_config`. If there is a better way to do it, I am open to it.

config/rfc3339.MD
mbohlool: True. This is, however, a release blocker and I don't want to wait for that contribution to go through. I suggest we proceed with this and either contribute to the that one or even release the one I copied as a pypi package for all to use (I have a feeling by looking at the code that this one is better, but we can decide on that later).

config/kube_config.py
mbohlool: to pass the config_file to _save_kube_config. If there is a better way to do it, I am open to it.

config/kube_config_test.py
mbohlool: Will fix this with some other tests.

config/rfc3339.MD
mbohlool: I've looked at different options and decided to implement my own. Some options are too restrictive for our use and the others are too admissive. I've implemented a simple balanced one in `dateutil.py`.

mbohlool:
@pokoli I am going to add some tests for dateutil but this is basically ready for review and I already tested it for GKE. Please take another look. I would like to merge this as soon as possible to cut a new release.

config/kube_config.py
pokoli: AFAIU it should work with: 
`
config_persister = _save_kube_config`

config/rfc3339.MD
pokoli: Ok, as it's a release blocker I think the unique option we have is to include the code in the client.

config/kube_config.py
mbohlool: `config_file` is a local variable. How should I pass it to `_save_kube_config`?

pokoli: Ok, I didn't understand at all the code.

Then you should probably use the [partial function from functools standard library. ](https://docs.python.org/2/library/functools.html#functools.partial)

This will make the code clearer.

mbohlool: I inline'ed `_save_kube_config`. it should be cleaner. is it good or do you still suggest partial function?

config/kube_config_test.py
mbohlool: Done.

config/kube_config.py
pokoli: Ok, now the usage it's much clearer.

mbohlool:
Thanks for the review @pokoli I think I addressed all of your comments and tests are passing. Can you take a final look please?

config/kube_config.py
```diff
@@ -308,21 +341,35 @@ def load_kube_config(config_file=None, context=None,
         from config file will be used.
     :param client_configuration: The kubernetes.client.ConfigurationObject to
         set configs to.
+    :param persist_config: IF True, config file will be updated when changed
```
pokoli: IF -> If

config/rfc3339.MD
mbohlool: For the record, I didn't copy code from any of those libraries. I've implemented it again by reading rfc. Of course my implementation can be influenced by those libraries, but I stick to the rules of rfc3339. None of the libraries I checked were like that, they were either relaxer or stricter. Also our implementation forces a UTC timezone.

config/kube_config_test.py
```diff
@@ -36,6 +38,10 @@ def _base64(string):
     return base64.encodestring(string.encode()).decode()
 
 
+def _raise_exception(st):
```
pokoli: Why not directly raising the exception?

pokoli:
@mbohlool for me you can merge once fixed.

config/kube_config_test.py
mbohlool: Why do you mean? I am passing this function s a function pointer to config loader and expect the function pointer (that suppose to update GCE token) never been called in the test.

pokoli: Ok, I didn't now that you can not raise inside lambda (always learning from codereview). 

Forget about it.

PR #28, opened 2017-08-20
fix: reset _stop when new stream is called

tomplus:
It resolves the issue: https://github.com/kubernetes-incubator/client-python/issues/326


mbohlool:
/lgtm

Thanks.

PR #29, opened 2017-08-22
Use a higher value for connecion_pool_maxsize.

### LINKED ISSUE #26: Consider setting connection_pool_maxsize to a higher value by default
Currently, it is set to 1. Which means the moment you start doing concurrent requests to the same kubernetes master, you're going to end up thrashing the connection pool & reducing performance.

Since the most likely case is that we're going to be making many requests to the same IP rather than many requests to different IP, setting this to a larger value by default will (and does!) improve performance.

pokoli:
Fixes #26

### DIFF
--- a/configuration.py
+++ b/configuration.py
@@ -20,6 +20,7 @@
 
 from __future__ import absolute_import
 
+import multiprocessing
 import logging
 import sys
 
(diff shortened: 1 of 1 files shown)


mbohlool:
This looks good to me. I will let @yuvipanda test and confirm this before merging. Thanks @pokoli.

configuration.py
```diff
@@ -83,11 +84,11 @@ def __init__(self):
...
         # urllib3 connection pool's maximum number of connections saved
-        # per pool. Increasing this is useful for cases when you are
-        # making a lot of possibly parallel requests to the same host,
-        # which is often the case here.
-        # When set to `None`, will default to whatever urllib3 uses
-        self.connection_pool_maxsize = None
+        # per pool. urllib3 uses 1 connection as default value, but this is
+        # not the best value when you are making a lot of possibly parallel
+        # requests to the same host, which is often the case here.
+        # cpu_count * 5 is used as default value to increase performance
```
yuvipanda: Add a note saying we're doing this because this is the default for ThreadPoolExecutor?

pokoli: Added, thanks.

pokoli:
Thanks for the review guys!

PR #30, opened 2017-08-25
Enable oidc with token refresh

dmyerscough:


mbohlool:
@dmyerscough  what happened to this PR? It looked like a good idea to support oidc.

PR #32, opened 2017-09-06
Add drain and cordon functions.

huseyinyilmaz:
Issue:
https://github.com/kubernetes-incubator/client-python/issues/188

Hello,

I worked on drain and cordon function prototypes and I wanted to get feedback before getting going further.

Can I get a code review on this. I will still need to write tests. Error handling seems to be OK since we are using other api's in the system.


mbohlool:
@huseyinyilmaz before you spend more time on this, I think we need to reach a decision/timeline for moving this functionality to server. It look like to me that is the best option. I would not have any objection to add this as beta to the client if the timeline for kubernetes/kubernetes#25625 is not working for us but need to be aware of that and even put our efforts there to fix this for all clients.

huseyinyilmaz:
@mbohlool, I was not aware that this functionality was being moved to server which seems like a better option than implementing the functionality for each client.I think we should close this pr for now. We can always reopen it if implementation on server is delayed.

pokoli:
Implementation on server is labeled as "help-wanted", so I'm not sure if anyone is working on it, but for sure it won't land on server till at least 1.9 (as 1.8 is already on beta).  

@huseyinyilmaz probably you will help to resolve the issue.

On the other hand, we can implement it here and remove when the support for the first version on which is implemented is dropped. But I don't like to much the idea to implement something that will be removed.

huseyinyilmaz:
Agreed. There is no need to repeat the the effort if it is going to be implemented anyway. Iw ill just close this and take a look at the issue on server.
Thanks @pokoli

PR #34, opened 2017-10-03
Change utility functions to use newly generated client

mbohlool:
Ref: https://github.com/kubernetes-incubator/client-python/issues/341


mbohlool:
We can ignore CI failure on this PR as the upstream client is updated to this change in https://github.com/kubernetes-incubator/client-python/pull/353 and if CI passes there, it would be good enough.

config/kube_config.py
```diff
@@ -355,10 +352,17 @@ def _save_kube_config(config_map):
...
+    loader = _get_kube_config_loader_for_yaml_file(
         config_file, active_context=context,
-        client_configuration=client_configuration,
-        config_persister=config_persister).load_and_set()
+        config_persister=config_persister)
+    if client_configuration is None:
+        config = type.__call__(Configuration)
+        loader.load_and_set(config)
+        Configuration.set_default(config)
+        print("ZZZ1: %s" % config.api_key)
```
pokoli: Print statement should be removed :)

config/kube_config.py
```diff
@@ -368,8 +372,8 @@ def new_client_from_config(
     """Loads configuration the same as load_kube_config but returns an ApiClient
     to be used with any API object. This will allow the caller to concurrently
     talk with multiple clusters."""
-    client_config = ConfigurationObject()
+    client_config = type.__call__(Configuration)
```
pokoli: Why is the __call__ required?

stream/ws_client.py
```diff
@@ -48,9 +48,10 @@ def __init__(self, configuration, url, headers):
         if headers and 'authorization' in headers:
             header.append("authorization: %s" % headers['authorization'])
 
-        if configuration.ws_streaming_protocol:
-            header.append("Sec-WebSocket-Protocol: %s" %
-                          configuration.ws_streaming_protocol)
+        if headers and 'sec-websocket-protocol' in headers:
+            header.append("sec-websocket-protocol: %s" % headers['authorization'])
```
pokoli: Not autorization but 'sec-websocket-protocol'

pokoli: Indeed it can be use get with default value: 

`headers.get('sec-websocket-procotol', 'v4.channel.k8s.io')`

config/kube_config.py
mbohlool: The PR is still a Work In Progress :)

config/kube_config.py
mbohlool: We need a fresh configuration object not the default one. With the change in Configuration class in swagger-codegen (that we did), creating a configuration object will give you the default object that has been set before. But here if we get that, we would have mixed old/new configuration. We rather want a fresh Configuration object with `__init__` default values not the `set_default` 'ed object.

stream/ws_client.py
mbohlool: doesn't help much because if headers is None we still want a default value.

mbohlool:
Now this is ready to review.

mbohlool:
@pokoli @dims Can you help reviewing this PR? Please ignore CI failure. That won't fix until I submit this and merge kubernetes-incubator/client-python#353

stream/ws_client.py
dims: Mehdy, we are checking for
```
'sec-websocket-protocol' in headers
```
BUT then using on line 52
```
headers['authorization']
```

dims: line 52 should be 
```
header.append("sec-websocket-protocol: %s" % headers['sec-websocket-protocol'])
```

mbohlool: Nice catch. Fixed.

dims:
LGTM 👍

PR #35, opened 2017-10-12
Fix exec command parameter expansion

mbohlool:
Simplify command parameter expansion of websocket_call that will also fix a bug with it (https://github.com/kubernetes-incubator/client-python/pull/346 originally tried to fix this before we move websocket code to the base repo). cc @eyeofthefrog

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -229,32 +229,21 @@ def websocket_call(configuration, *args, **kwargs):
     apiClient.request method."""
 
     url = args[1]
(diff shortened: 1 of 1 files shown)


dims:
LGTM 👍 we could add a test case when possible

mbohlool:
Thanks @dims. I spend a little time on writing a test but I needed to mock WebSocket class and I didn't find an easy way to do it. I will merge this without a test for now.

PR #36, opened 2017-10-17
Keep the watch action working forever

lichen2013:
Fixes issue: https://github.com/kubernetes-incubator/client-python/issues/124

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -63,6 +63,7 @@ class Watch(object):
         self._raw_return_type = return_type
         self._stop = False
         self._api_client = client.ApiClient()
+        self.resource_version = 0
 
     def stop(self):
         self._stop = True
@@ -81,15 +82,15 @@ class Watch(object):
         if return_type:
             obj = SimpleNamespace(data=json.dumps(js['raw_object']))
             js['object'] = self._api_client.deserialize(obj, return_type)
+            if hasattr(js['object'], 'metadata'):
+                self.resource_version = js['object'].metadata.resource_version
         return js
 
-    def stream(self, func, keep=False, *args, **kwargs):
+    def stream(self, func, *args, **kwargs):
         """Watch an API resource and stream the result back via a generator.
 
         :param func: The API function pointer. Any parameter to the function
                      can be passed after this parameter.
-
-        :param keep: Flag to keep the watch work all the time.
 
         :return: Event object with these keys:
                    'type': The type of event such as "ADDED", "DELETED", etc.
@@ -116,6 +117,7 @@ class Watch(object):
         kwargs['watch'] = True
         kwargs['_preload_content'] = False
 
+        timeouts = ('timeout_seconds' in kwargs)
         while True:
             resp = func(*args, **kwargs)
             try:
@@ -124,8 +126,9 @@ class Watch(object):
                     if self._stop:
                         break
             finally:
+                kwargs['resource_version'] = self.resource_version
                 resp.close()
                 resp.release_conn()
 
-            if not keep or self._stop:
+            if timeouts or self._stop:
                 break
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -83,11 +83,13 @@ class Watch(object):
             js['object'] = self._api_client.deserialize(obj, return_type)
         return js
 
-    def stream(self, func, *args, **kwargs):
+    def stream(self, func, keep=False, *args, **kwargs):
         """Watch an API resource and stream the result back via a generator.
 
         :param func: The API function pointer. Any parameter to the function
                      can be passed after this parameter.
+
+        :param keep: Flag to keep the watch work all the time.
 
         :return: Event object with these keys:
                    'type': The type of event such as "ADDED", "DELETED", etc.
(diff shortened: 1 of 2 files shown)


watch/watch.py
```diff
@@ -113,12 +115,17 @@ def stream(self, func, *args, **kwargs):
...
+            try:
+                for line in iter_resp_lines(resp):
+                    yield self.unmarshal_event(line, return_type)
+                    if self._stop:
+                        break
+            finally:
+                resp.close()
+                resp.release_conn()
+
+            if not keep or self._stop:
```
mbohlool: This will result in all cached event to be delivered again. If we want to reconnect, we should catch the last resourceVersion and pass it to the watch call to make sure we continue receiving new events not old cached events. Also would be nice if we can fix this at root too (find the reason for timeout in http client and fix that). I think the server keeps the connection alive and it is the client that has a pre-defined timeout somewhere.

lichen2013: Thanks, let me do some changes to add the resourceVersion.

I'm not very sure, but someone do told me the connection is cancelled by server.
I did some simple checked in client-go:

client-go/tools/cache/reflector.go

    // Run starts a watch and handles watch events. Will restart the watch if it is closed.
    // Run will exit when stopCh is closed.
    func (r *Reflector) Run(stopCh <-chan struct{}) {
           glog.V(3).Infof("Starting reflector %v (%s) from %s", r.expectedType, r.resyncPeriod, r.name)
           wait.Until(func() {
                   if err := r.ListAndWatch(stopCh); err != nil {
                           utilruntime.HandleError(err)
                   }
           }, r.period, stopCh)
    }

 apimachinery/pkg/util/wait/wait.go

    // Until loops until stop channel is closed, running f every period.
    //
    // Until is syntactic sugar on top of JitterUntil with zero jitter factor and
    // with sliding = true (which means the timer for period starts after the f
    // completes).
    func Until(f func(), period time.Duration, stopCh <-chan struct{}) {
            JitterUntil(f, period, 0.0, true, stopCh)
    }

So, I think it should be ok to just reconnect. 
And, I will try do more checks to find/confirm the root cause.  :)

lichen2013: Thanks!
Let me do a little more checks.

karmab: for what it's worth, i indeed got this same kind of issue solved using @mbohlool recommendation and this kind of snippet:

```
    resource_version = None
    while True:
        if resource_version is None:
            stream = watch.Watch().stream(crds.list_cluster_custom_object, DOMAIN, "v1", "guitars")
        else:
            stream = watch.Watch().stream(crds.list_cluster_custom_object, DOMAIN, "v1", "guitars", resource_version=resource_version)
        for event in stream:
            obj = event["object"]
```

lichen2013: Some updates for the timeout setting:

There are two inputs available in the client can set timeouts:

a) timeout_seconds:  This value will not used by client but included in the URL: 

    https://localhost:6443/api/v1/namespaces/default/pods?labelSelector=app%3Ddemo&timeoutSeconds=100&watch=True.

So, this setting is actually used by server.

b) _request_timeout 

This can accept 2 type inputs, int or tuple with length equals to 2, checked at 

    kubernetes.client.rest.RESTClientObject.request()

The value will eventually set to the socket used for the connection.
b.1) if the input is tuple, the first value will actually been ignored.
b.2) when timeout happens, an exception will be raised, for example:

    urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='localhost', port=6443): Read timed out.

b.3) when the value is not set,  the default value is None, socket has no timeout.
    
    https://docs.python.org/2/library/socket.html#socket.getdefaulttimeout



Haven't figure out where/how server sets timeout, but I believe when we leave both timeout settings empty, client will keeps the connection alive.

mbohlool: Nice update on the timeout. `timeout_seconds` is what server expect for watch calls and if you don't set it, it won't timeout the connection. There is a hidden timeout on the request that fortunately the generated client don't know about and cannot set.

Even with not setting the timeout, the connection can fail or network can fail and it is a good idea to have this flag if we make it smart enough to continue the watch using `resource_version`.

lichen2013: Could you explain a little more about the "There is a hidden timeout on the request that fortunately the generated client don't know about and cannot set".  Really courious about this....

lichen2013: Continue the watch using resource_version updated.
Also, I did a little change. Since we already have "timeout_seconds" to let user set the timeouts, when the value is empty, I suppose this mean watch forever, no extra flag is needed.

watch/watch.py
```diff
@@ -113,12 +116,19 @@ def stream(self, func, *args, **kwargs):
...
+                for line in iter_resp_lines(resp):
+                    yield self.unmarshal_event(line, return_type)
+                    if self._stop:
+                        break
+            finally:
+                kwargs['resource_version'] = self.resource_version
+                resp.close()
+                resp.release_conn()
+
+            if timeouts or self._stop:
```
mbohlool: What if the user set timeouts but it has not been passed yet?

mbohlool:
@lichen2013 This looks very good. Only one comment, otherwise good to go.

watch/watch.py
lichen2013: I'm still trying to find the proof that the exit happens we saw at the beginning is caused by the client side.

I do find the flag 'min-request-timeout' in doc [kube-apiserver](https://github.com/kubernetes/kubernetes/blob/release-1.1/docs/admin/kube-apiserver.md), and it is said that the apiserver actually has a timeout setting by itself: 

> --min-request-timeout=1800: An optional field indicating the minimum number of seconds a handler must keep a request open before timing it out. Currently only honored by the watch request handler, **which picks a randomized value above this number as the connection timeout, to spread out load**.

And if it is not set,  the value for it is 1800:

>https://github.com/kubernetes/apiserver/blob/master/pkg/server/config.go#L248

    MinRequestTimeout: 1800,

So, base on the what I know now: I don't think timeout will happen before the timeouts setting.

mbohlool:
@lichen2013 Only remaining item here is to reconnect if the user specified timeout is not passed yet (in addition to the case of not setting timeout which reconnect the watch always). I would want to include this in 4.0 client but that needs to be done soon.

PR #38, opened 2017-11-03
Check if the cached temp conf file still exists before using it.

amitsrivastava:
On some systems older unused temp files get cleaned up regularly. So,
we cannot rely on the availability of the temp_files in long running
processes.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -49,7 +49,7 @@ def _create_temp_file_with_content(content):
     # Because we may change context several times, try to remember files we
     # created and reuse them at a small memory cost.
     content_key = str(content)
-    if content_key in _temp_files:
+    if content_key in _temp_files and os.path.isfile(_temp_files[content_key]):
         return _temp_files[content_key]
     _, name = tempfile.mkstemp()
     _temp_files[content_key] = name


config/kube_config.py
```diff
@@ -49,7 +49,7 @@ def _create_temp_file_with_content(content):
     # Because we may change context several times, try to remember files we
     # created and reuse them at a small memory cost.
     content_key = str(content)
-    if content_key in _temp_files:
+    if content_key in _temp_files and os.path.isfile(_temp_files[content_key]):
```
arunmk: There is still a small window for a race here: after the check of is_file, the file may be deleted (before line 53), or the file may be deleted after the function returns, but before the file is consumed. There should be a more generic solution to this, or the tracking of the file should be delegated to the caller.

amitsrivastava: That's absolutely correct @arunmk. But, what you suggest is a way bigger change. The temp file names are being passed all over and beyond the scope of this module (kube_config), so the changes have to go all the way up.

amitsrivastava: Currently, the situation is such that once you get into a bad condition, you can never get out without a restart of the python process. But, with this change, most such cases should get handled and even the one that you pointed can only cause a one-time exception.

arunmk: If we use the $TMPDIR, I don't see a way out that's better than the solution that you have here @amitsrivastava . Could you also add an API documentation change that says something to the effect of: 'The file that the API emits may not exist due to cleanup policy of $TMPDIR. In that case, the pattern is to retry the API until the file exists." That will clarify the usage of the API.

tomplus: Temporary files can be safe deleted only if these files are not accessed by processes for a long time. For example `tmpwatch` (popular on Red Hat) works in this way. Because of this I suggest updating mtime to prevent deleting files. It also removes race condition.

```python
 if content_key in _temp_files:
    try:
       os.utime(_temp_files[content_key], None)
       return _temp_files[content_key]   # file exists and mtime has been updated
    except OSError as err:
       if err.errno == errno.ENOENT:
           # ups, file disappeared, has to be recreated
           pass
       else:
           raise 
```

How does it look ?

arunmk: @tomplus it may work but it's very implementation (Linux flavor and version) dependent and need not be future-proof. Also users may just delete the content of $TMPDIR at any point of time. The guarantees of $TMPDIR are not much: any files kept open will not be deleted due to underlying filesystem semantics while the rest of files could be reclaimed at any point.

This fix may also not always work, because the caller of the function need not immediately use the content of the file. Suppose there's a pathological scenario where the content will be used 2s later while the cleanup time is set at 1s. We'll fall into the same issue in that case.

Hence my thoughts are to fail cleanly and early and document the behavior rather than introducing more complexity in the code.

That said, this change does not add any new failure modes and covers a large number of cases, so if it fits into the theme of this codebase it's fine with me.

arunmk: @amitsrivastava @tomplus it looks like the /run directory is better for such purposes. From http://www.h-online.com/open/news/item/Linux-distributions-to-include-run-directory-1219006.html:
"
On the Fedora project's developer list, systemd developer Lennart Poettering has announced the introduction of a /run directory in the root directory and provided detailed background explanations. Similar to the existing /var/run/ directory, the new directory is designed to allow applications to store the data they require in order to operate. This includes process IDs, socket information, lock files and other data which is required at run-time but can't be stored in /tmp/ because programs such as tmpwatch could potentially delete it from there.
"

So if feasible, the /var/run or the /run directory may be a better location to create these temporary files.

mbohlool: Is keeping the temp file open a bad idea?

scottilee:
Hey @amitsrivastava, any update on this? I can get someone to review it afterwards.

Since this PR has been open for a while, if it is no longer valid please close it.

arunmk:
@scottilee @amitsrivastava I can take this forward and create a patch if you are fine by it

pigletfly:
any updates on this?

ceroloy:
I am seeing several PRs are linked here as related and some of them as merged. Is this issue resolved ? If so can you please share in which version is this fixed ?

Thank you

config/kube_config.py
sn003: Hi, Any update on this issue? Looks like it is not merged. How to address this issue ?

reachnatp: I am facing the same issue as well, what would be the best way to work around?

onecer: I choose store temp in another directory.

PR #40, opened 2018-01-01
Fix issue when refreshing service account token

jdartigalongue:
@mbohlool  I believe this will fix the below issues:
https://github.com/kubernetes-incubator/client-python/issues/339
https://github.com/kubernetes-incubator/client-python/issues/386

I was able to reproduce this issue by either setting GOOGLE_APPLICATION_CREDENTIALS to a service account json key file or by setting the gcloud default application login to a service account. 

Unlike user accounts, it seems that service accounts require that we set the scope.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -129,7 +129,9 @@ class KubeConfigLoader(object):
         self._config_persister = config_persister
 
         def _refresh_credentials():
-            credentials, project_id = google.auth.default()
+            credentials, project_id = google.auth.default(
+                scopes=['https://www.googleapis.com/auth/cloud-platform']
+            )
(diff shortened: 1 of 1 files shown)


jdartigalongue:
Bum to verify CLA

mbohlool:
Thanks for your contribution. You also need to send a PR to the main repo updating this sub-module.

michrabi:
how can I get this fix? I tried cloning and I don't see it. 
Thanks

pokoli:
@michrabi you should create a PR to the main repo (https://github.com/kubernetes-client/python) updating the submodule to include this commit. See: https://github.com/kubernetes-client/python/blob/master/devel/submodules.md

Thanks

PR #41, opened 2018-01-05
id/bearer token auth support for API client configuration

jroberts235:
Support for use of id/bearer token in kube.conf when auth'ing with kubernetes API

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -171,11 +171,25 @@ class KubeConfigLoader(object):
         """
         if not self._user:
             return
+        if self._load_id_token():
+            return
         if self._load_gcp_token():
             return
         if self._load_user_token():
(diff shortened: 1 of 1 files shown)


mbohlool:
Please sign CLA and also prove a provider example that uses this.

grillz:
CoreOS has a tool that writes the kubecfg this way. Regardless its a valid kubecfg so the client should be able to handle it.

jroberts235:
CLA signed.

mbohlool:
signature is not registered yet. Please try again.

jroberts235:
CLA was signed Jan 19.

mbohlool:
ref: https://github.com/kubernetes/community/blob/master/CLA.md

PR #42, opened 2018-01-23
Add load kube config from dict

pigletfly:
sometimes the kubeconfig is not stored in a file,may be in db or something.So I think we need a method to load kubeconfig from dict, like client-go https://godoc.org/k8s.io/client-go/tools/clientcmd#RESTConfigFromKubeConfig

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -313,12 +313,15 @@ class ConfigNode(object):
             'Expected object with name %s in %s list' % (name, self.name))
 
 
-def _get_kube_config_loader_for_yaml_file(filename, **kwargs):
-    with open(filename) as f:
-        return KubeConfigLoader(
-            config_dict=yaml.load(f),
-            config_base_path=os.path.abspath(os.path.dirname(filename)),
-            **kwargs)
+def _get_kube_config_loader(config_file=None, config_dict=None, **kwargs):
+    if config_dict:
+        return KubeConfigLoader(config_dict=config_dict, **kwargs)
+    if config_file:
+        with open(config_file) as f:
+            return KubeConfigLoader(
+                config_dict=yaml.load(f),
+                config_base_path=os.path.abspath(os.path.dirname(config_file)),
+                **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


config/__init__.py
```diff
@@ -15,4 +15,4 @@
 from .config_exception import ConfigException
 from .incluster_config import load_incluster_config
 from .kube_config import (list_kube_config_contexts, load_kube_config,
-                          new_client_from_config)
+                          new_client_from_config, load_kube_config_from_dict)
```
roycaihw: Please see the travis ci log for formatting errors

config/kube_config.py
```diff
@@ -321,6 +321,10 @@ def _get_kube_config_loader_for_yaml_file(filename, **kwargs):
             **kwargs)
 
 
+def _get_kube_config_load_for_dict(config_dict, **kwargs):
+    return KubeConfigLoader(config_dict=config_dict, **kwargs)
```
roycaihw: You don't need to wrap the `KubeConfigLoader` IMO

config/kube_config.py
```diff
@@ -366,6 +370,18 @@ def _save_kube_config(config_map):
         loader.load_and_set(client_configuration)
 
 
+def load_kube_config_from_dict(config_dict=None, context=None,
+                               client_configuration=None):
+    loader = _get_kube_config_load_for_dict(config_dict=config_dict,
+                                            active_context=context)
+    if client_configuration is None:
```
roycaihw: Please make this line and below into a `util(loader, client_configuration)` method that can be used by both `load_kube_config` and `load_kube_config_from_dict`

config/kube_config.py
pigletfly: Can we add a `config_dict` in `new_client_from_config` `load_kube_config`, and then make a `_get_kube_config_loader(config_file, config_dict, active_context, config_persister)`.Then use `config_dict` or `config_file` to init `KubeConfigLoader`

pigletfly:
ping @roycaihw

PR #31, opened 2017-08-25
Enable oidc auth

dmyerscough:

PR #48, opened 2018-02-23
Add oidc auth 

ltamaster:


dmyerscough:
Hi @ltamaster you will want to run the following against the files you modified:-

```bash
$ pip install pycodestyle
$ pycodestyle ...
```

The Travis CI is failing due to coding style errors.

ltamaster:
@dmyerscough thanks, I will try

mcalmer:
JFYI: I tried this patch and it works fine against our kubernetes cluster. Seems that only one unit test is failing. Maybe the last thing what prevents this PR from being merged.

dmyerscough:
Nice work @ltamaster, I just saw you message now and see you applied a fix to the unit test.

ltamaster:
thanks @dmyerscough, finally I found the issue !!!

mvle:
Re-posting comment here from closed PR #31. Seems to affect this new PR as well.

I apologize in advance if this is already supported, but from looking at the PR, I think there needs to be a way to refresh the token not only at configuration initialization time but also during invocation of the K8 API calls. The token might expire right before a call thus causing the call to fail with an authentication failure. This will be needed for services that run a long time and rely on being able to reach the K8 Api Server (also short services that happen to be unlucky).

I think this can be done with a check to 'exp' field in the id-token right before a call_api or catch a 401 return status and perform a refresh token operation and update the configuration object with the new id_token and refresh_token values from the OAuth server.

mcalmer:
@roycaihw - what prevents this pr to get merged and added to 6.0.0?

roycaihw:
> I think this can be done with a check to 'exp' field in the id-token right before a call_api or catch a 401 return status and perform a refresh token operation and update the configuration object with the new id_token and refresh_token values from the OAuth server.

@mvle I agree that we need to ensure `id_token` (and also `access_token` for gcp) is valid and refresh as needed. However, currently `swagger-codegen` doesn't support [refreshing token](https://github.com/swagger-api/swagger-codegen/issues/4000) of OAuth2. Without upstream change, the generated client's support for refreshing token would be hacky. cc @yliaog for more insight here

> what prevents this pr to get merged and added to 6.0.0?

@mcalmer Sorry I didn't get a chance to look at this yet. `client-python` is currently 6.0.0b1. This pr can still get in 6.0.0 which will get released after k8s 1.10 rolls out.

yliaog:
i think the suggestion for refreshing token is fine, but can be done in a separate PR. 

For this PR, I see many commits, @ltamaster could you please squash the commits?

ltamaster:
Hi @yliaog I can do that, but I understand that I need a new PR for that right?

yliaog:
i think you can use the same PR for squashing commits, so all the history can still be kept.

nurus:
I've been testing OIDC for authentication using the work done in this PR and came across an issue refreshing credentials because I do not specify `idp-certificate-authority-data` in my kubeconfig. This issue has already been brought up: https://github.com/kubernetes-client/python/issues/368#issuecomment-337778192. 

I don't think there should be an expectation to specify this key if you're using an IdP which uses a CA found in a trusted CA bundle. The request to the IdP's openid-configuration endpoint using the rest client falls back to the [Mozilla CA bundle if a certificate isn't specified](https://github.com/kubernetes-client/python/blob/369366a813964e3d72e91d402c5f471275e7a3e5/kubernetes/client/rest.py#L72).

Another potential issue is the lack of support for the `idp-certificate-authority` key which allows you to load CA certificates from a file. I was attempting to use this to get around the issue mentioned above by loading a trusted root CA bundle. I could convert the entire file to base64 and add it as a value for `idp-certificate-authority-data` but this results in an unwieldy kubeconfig file.

ltamaster:
Hi @yliaog, I tried to do a squash with a rebase, but It seems it created new commits on the PR. I think I need to create a new PR, what do you think?

yliaog:
not sure how you did it, did you try something like what's described here (https://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit)

ltamaster:
Hi @yliaog, finally I squashed the commits, please take a look.
Thanks
Luis

yliaog:
thanks @ltamaster, the remaining issues are tracked separately.

yliaog:
@mvle could you please file another issue to track the feature request for refreshing token?

mvle:
@yliaog done. kubernetes-client/python#492

ltamaster:
Hi @yliaog 

I noticed this PR is merged, but it hasn't been merged on the client [repo](https://github.com/kubernetes-client/python). 

Are you waiting for the other issues or do I need something else?

Thanks
Luis

roycaihw:
@ltamaster I was planning on updating the base submodule when we do the 6.0.0 release. PR is also welcome to do the update :) (e.g. https://github.com/kubernetes-client/python/pull/475)

PR #49, opened 2018-03-03
allow watch constructor to take a apiclient object as a parameter

huang195:
To be consistent with other constructors that allow ApiClient to be passed as a parameter.


huang195:
@mbohlool ptal

PR #52, opened 2018-03-21
binary websocket streams

aogier:
Hi, this PR gives `bytes` streams first class citizenship in this codebase, unlocking future features like binary websocket streams as per kubernetes-client/python#476 (for which I'm going to submit code against python client for this).

I hope this could be useful, I've worked unobtrusively so nothing existing should break !

Thank you, regards

### DIFF
--- a/run_tox.sh
+++ b/run_tox.sh
@@ -34,7 +34,7 @@ SCRIPT_ROOT=`pwd`
 popd > /dev/null
 
 cd "${TMP_DIR}"
-git clone https://github.com/kubernetes-client/python.git
+git clone https://github.com/aogier/k8s-client-python.git -b feature/476-copy-example python
 cd python
 git config user.email "kubernetes-client@k8s.com"
 git config user.name "kubenetes client"
--- a/run_tox.sh
+++ b/run_tox.sh
@@ -34,7 +34,7 @@ SCRIPT_ROOT=`pwd`
 popd > /dev/null
 
 cd "${TMP_DIR}"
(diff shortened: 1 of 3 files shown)


aogier:
Ok this night I've made some progresses, what I'm discovered is if I patch corev1api [this way](https://github.com/kubernetes-client/python/compare/kubernetes-client:369366a...aogier:ee6dbe5#diff-6385abebebf123ea92c86e2bf523f687L929) I can modify ws_client so it works with [bytes and not strings](https://github.com/aogier/k8s-client-python-base/compare/11da619c9ef7b2e26fe162e5feeb72d5bedcc5c9...a7dc7402528afe5e17ca8cc309734e6c91225d0c#diff-eb24b95d54baa73740535635b5f56b5a).

Problems:
 1. this is a breaking change, as tools using stream() on python3 must now decode non-binary data
 1. corev1api is swagger generated se we should work swagger side, unfortunately I have no idea on how to do it ATM and I have to study it
 1. python3.5 functional tests fail because before 3.6 `json.loads` can't handle bytes, only strings. I'm evaluating a backport proposal in python bug tracker

This is proving to be more complex than I thought at the beginning, but I'd really like to have this feature implemented, what do you think ? :)

Thank you, regards

ltamaster:
Hey guys, any news about this PR?

zhaofengli:
I have tested the change, and it appears to work well for my use cases. I'd really like this merged back upstream.

Ananasr:
any news ?

PR #55, opened 2018-03-30
fixing flipped sign in expiry time padding

flylo:
We ran into a bug where any job that runs long enough for the API token to expire will fail with a `401 Unauthorized ApiException`. We found that this was due to a flipped sign in the padding added to the expiry. I've updated the `_is_expired` method to return `True` if the token expires _in 5 minutes_, as opposed to the current logic which returns `True` if the token expired _5 minutes ago_. I've also added an extra assertion to ensure that the `_load_gcp_token` method actually updates the expiry timestamp.

Please let me know if you have any questions or comments!

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -63,7 +63,7 @@ def _create_temp_file_with_content(content):
 
 
 def _is_expired(expiry):
-    return ((parse_rfc3339(expiry) + EXPIRY_SKEW_PREVENTION_DELAY) <=
+    return ((parse_rfc3339(expiry) - EXPIRY_SKEW_PREVENTION_DELAY) <=
             datetime.datetime.utcnow().replace(tzinfo=UTC))
 
 
--- a/config/kube_config_test.py
+++ b/config/kube_config_test.py
@@ -22,10 +22,9 @@ import unittest
 
 import mock
 import yaml
-from six import PY3
+from six import PY3, next
 
 from .config_exception import ConfigException
-from .dateutil import parse_rfc3339
 from .kube_config import (ConfigNode, FileOrData, KubeConfigLoader,
                           _cleanup_temp_files, _create_temp_file_with_content,
                           list_kube_config_contexts, load_kube_config,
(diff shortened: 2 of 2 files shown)


flylo:
Looks like i'm getting a PEP8 issue in python 2.7 but I can't tell what it is. Can someone help me interpret those log messages?

flylo:
also FYI this stems from a discussion in:
https://github.com/kubernetes-client/python/issues/492

flylo:
anyone have any idea what's wrong with my py2.7 build? it doesn't say what the actual failure is.

flylo:
cc @mbohlool who seems to be the one who worked on the expiry time logic

yliaog:
@flylo I reran the CI, and it is still failing https://travis-ci.org/kubernetes-client/python-base/jobs/361191252

in particular, i see the following error message.
/tmp/tmp.ztIlFmpO08/python/kubernetes/config/kube_config_test.py:215:80: E501 line too long (80 > 79 characters)

flylo:
@yliaog still won't pass and all I can see are a bunch of user warnings to use pycodestyle instead of pep8. what exactly does this mean?
`ERROR: InvocationError for command '/tmp/tmp.mcnTCgxokv/python/scripts/update-pep8.sh' (exited with code 1)`

flylo:
@yliaog what is the purpose of this script [here](https://github.com/kubernetes-client/python/blob/master/scripts/update-pep8.sh#L80)? Why does my build fail because some env is set or my git status is clean?

yliaog:
I'm not quite sure, I can take a look tomorrow.

 https://github.com/kubernetes-client/python/blame/master/scripts/update-pep8.sh#L80 shows @dims added it in https://github.com/kubernetes-client/python/commit/5c823098409695fc2c7525cd1416fad68c6e83e2

flylo:
Figured out that the diff is between the PR files and the pep8 auto-formatted files. There should probably be some log messages in that script as it wasn't clear to me what that was doing until I went back to the PR where it was merged.
@yliaog this good to merged now?

yliaog:
Yes, thanks for fixing it.

PR #58, opened 2018-04-13
add environment INSECURE_SKIP_TLS_VERIFY

jzwlqx:
In the scenario a two-tier PKI hierarchy is used, the CA of kubernetes cluster is a sub-CA that signed by the root CA. The CA file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt in the kubernetes pod now is a sub-CA but not the root CA. In some versions of openssl, a sub-CA can not be used as a trust anchor, `curl --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt https://kubernetes` will return an error "unable to get issuer certificate". The below example also will report an error "certificate verify failed"

```
from kubernetes import client, config
import os

# Configs can be set in Configuration class directly or using helper utility
print(os.environ)
config.load_incluster_config()

v1 = client.CoreV1Api()
print("Listing pods with their IPs:")
ret = v1.list_pod_for_all_namespaces(watch=False)
for i in ret.items:
        print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))
```

In this pull request, An environment INSECURE_SKIP_TLS_VERIFY is added, if set to "true", the server certificate verify will be skipped.

### DIFF
--- a/config/incluster_config.py
+++ b/config/incluster_config.py
@@ -76,9 +76,14 @@ class InClusterConfigLoader(object):
 
         self.ssl_ca_cert = self._cert_filename
 
+        self.verify_ssl = True
+        if os.environ.get("INSECURE_SKIP_TLS_VERIFY") == "true":
+            self.verify_ssl = False
+
     def _set_config(self):
         configuration = Configuration()
         configuration.host = self.host
+        configuration.verify_ssl = self.verify_ssl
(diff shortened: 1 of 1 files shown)


jzwlqx:
I have already signed the individual CLA, but the status here was not change. I don't know what to do next.

georgedriver:
Any update on this merge? We really want to this.

PR #67, opened 2018-05-22
Add '==' to base64 decode for py3

bcatubig:
Fixes https://github.com/kubernetes-client/python/issues/525


hanikesn:
The added padding should really depend on the lenght of the base64 string. See #70.

bcatubig:
Thanks @hanikesn, will close this

PR #69, opened 2018-05-25
remove required idp-certificate-authority-data in kubeconfig for oidc…

mvle:
… toke refresh, kubernetes-client/python#493


mvle:
I got myself added to the corporate CLA. How do I get git to perform a re-check?

hanikesn:
What about:
>Why would you not verify the certificate just because it isn't specified?

Issue #59, opened 2018-04-16
GCP Token is not refreshed


mikedanese:
The GCP token expires after 1 hour and is not refreshed by the transport.

mikedanese:
cc @kashomon

amj:
/subscribe

i'm also having this problem :(

dekkagaijin:
I can take this on

amj:
woohoo!

richardsliu:
I am having the same problem. Is there a fix planned for this?

richardsliu:
Also what is the recommended workaround? The only way I can think of is catching 401/403s on the client side, then calling gcloud get-credentials and retry. Is there a better way?

roycaihw:
@dekkagaijin has a fix that mitigates the problem during config loading https://github.com/kubernetes-client/python-base/pull/72

The main problem is that the client only check/refresh token during config loading-- it doesn't refresh token during api invocation. The problem applies to all authentication methods. Because the client is auto-generated by swagger-codegen, we cannot directly modify it. Suggested fixes are:

1. wrap the client to refresh token during api invocation (WIP in https://github.com/kubernetes-client/python/issues/492)
2. have another thread to refresh the token periodically in the background
3. maybe port a fix to upstream code generator somehow

richardsliu:
I'm doing this as a workaround:

```python
from kubernetes import client as k8s_client
from kubernetes import config as k8s_config
from kubernetes.client import rest

...
while retry_condition:
  try:
    retry_condition = False
    crd_api = k8s_client.CustomObjectsApi(client)
    # This will throw an ApiException with status 401 if the k8s client was 
    # instantiated over an hour ago
    return crd_api.get_namespaced_custom_object(...)
  except rest.ApiException as e:
    logging.exception("ApiException: %s", e)
    if e.status == 401 or e.status == 403:
      # 1. Reload the kube config, which will refresh GCP token;
      k8s_config.load_kube_config()
      # 2. Re-create the k8s client using the refreshed config.
      client = k8s_client.ApiClient()
      # Trigger retry of this block
      retry_condition = True
```

But it would be great to have this fixed in the client somehow.

TrevorEdwards:
It looks like the token is set on the client_configuration in one place:
https://github.com/kubernetes-client/python-base/blob/master/config/kube_config.py#L395-L402

[...]

PR #72, opened 2018-07-03
refresh GCP tokens if <55 mins of life left

dekkagaijin:
Partially mitigates #59 

Signed-off-by: Jake Sanders <jsand@google.com>

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -32,7 +32,8 @@ from kubernetes.client import ApiClient, Configuration
 from .config_exception import ConfigException
 from .dateutil import UTC, format_rfc3339, parse_rfc3339
 
-EXPIRY_SKEW_PREVENTION_DELAY = datetime.timedelta(minutes=5)
+EXPIRY_TIME_SKEW = datetime.timedelta(minutes=5)
+MINIMUM_GCP_TOKEN_TIME_REMAINING = datetime.timedelta(minutes=55)
 KUBE_CONFIG_DEFAULT_LOCATION = os.environ.get('KUBECONFIG', '~/.kube/config')
 _temp_files = {}
 
@@ -62,8 +63,16 @@ def _create_temp_file_with_content(content):
     return name
 
 
+def _is_stale(expiry):
+    return _has_min_lifespan(expiry, MINIMUM_GCP_TOKEN_TIME_REMAINING)
+
+
 def _is_expired(expiry):
-    return ((parse_rfc3339(expiry) - EXPIRY_SKEW_PREVENTION_DELAY) <=
+    return _has_min_lifespan(expiry, EXPIRY_TIME_SKEW)
+
+
+def _has_min_lifespan(expiry, min_lifespan):
+    return ((parse_rfc3339(expiry) - min_lifespan) <=
             datetime.datetime.utcnow().replace(tzinfo=UTC))
 
 
@@ -198,7 +207,7 @@ class KubeConfigLoader(object):
         if (('config' not in provider) or
                 ('access-token' not in provider['config']) or
                 ('expiry' in provider['config'] and
-                 _is_expired(provider['config']['expiry']))):
+                 _is_stale(provider['config']['expiry']))):
             # token is not available or expired, refresh it
             self._refresh_gcp_token()
 


dekkagaijin:
PTAL @mikedanese

config/kube_config.py
```diff
@@ -32,7 +32,7 @@
 from .config_exception import ConfigException
 from .dateutil import UTC, format_rfc3339, parse_rfc3339
 
-EXPIRY_SKEW_PREVENTION_DELAY = datetime.timedelta(minutes=5)
+MINIMUM_TOKEN_TIME_REMAINING = datetime.timedelta(minutes=55)
```
mikedanese: What does the metadata server do? If the metadata server also caches tokens, but refreshes them 10 seconds before expiry, then will this call the metadata server on every request for about 5 minutes?

Also can you explain the bug since I'm a python noob?

mikedanese:
cc @mbohlool @yliaog

config/kube_config.py
mikedanese: accounts.google.com provisions tokens with a max validity duration of 1 hour. I think the GCE metadata server provisions tokens valid for 30 minutes.

yliaog: i think this would cause constant token refreshing if the token validity life is <55 minutes.

dekkagaijin: > What does the metadata server do? If the metadata server also caches tokens, but refreshes them 10 seconds before expiry, then will this call the metadata server on every request for about 55 minutes?

@mikedanese IIRC the metadata server lazily mints new tokens when they've expired, but I'm definitely not an expert

> Also can you explain the bug since I'm a python noob?

Not really a 'bug' per se, it's just that long-running operations can very easily result in 401s/403s due to expired tokens. For example, by default, dockerd only pulls 3 image layers in parallel, meaning that especially large images being downloaded over real-world networks can fail half-way through.

> i think this would cause constant token refreshing if the token validity life is <55 minutes. also this python client is for generic use, not limited to GCP.

@yliaog  So only apply this minimum freshness to the GCP tokens? Works for me.

yliaog: I guess you can make it a parameter, instead of a const. so you can set it to the most appropriate value for your use case (for GCP?)

dekkagaijin: @yliaog It doesn't need to be shared between GCP and other credentials, I don't think, since the 1h lifespan on GCP tokens is unlikely to change. If the existing logic suffices for other credentials, I can limit this change to GCP credentials.

tomplus:
This doesn't solve problem with long-living applications. I suggest to use a thread to refresh the token periodically in the background.

config/kube_config.py
dekkagaijin: also, it seems like clients of this lib won't necessarily know a-priori which token source will be used: https://github.com/kubernetes-client/python-base/blob/master/config/kube_config.py#L179

Seems like this issue might be sufficiently addressed by limiting the change to the GCP token logic. There might be a separate task of factoring out the Authenticator logic into a plugin interface.

dekkagaijin:
@tomplus this isn't intended to completely solve the problem, just reduce the likelihood that it will be encountered.

> I suggest to use a thread to refresh the token periodically in the background.

What would be the periodicity?

Would automatically refreshing & retrying on 401/403 be insufficient?

config/kube_config.py
mikedanese: This method won't work well for the GCE compute metadata source where there is an extra layer of caching. This will result in a call to the metadata server per request.

dekkagaijin: > This method won't work well for the GCE compute metadata source where there is an extra layer of caching. This will result in a call to the metadata server per request.

Doesn't seem like it's possible to fully mitigate the issue, then, since AFAIK we can't coerce a refresh. We'd just have to hope that one retry is sufficient to pick up a token with enough lifespan left to complete a request.
From the docs at https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#applications:
> The metadata server caches access tokens until they have 60 seconds of remaining time before they expire.


Looks like the only 'good' mitigation would be to reload the kube-config and retry requests on 401/403?

config/kube_config.py
```diff
@@ -32,7 +32,8 @@
 from .config_exception import ConfigException
 from .dateutil import UTC, format_rfc3339, parse_rfc3339
 
-EXPIRY_SKEW_PREVENTION_DELAY = datetime.timedelta(minutes=5)
+EXPIRY_TIME_SKEW = datetime.timedelta(minutes=5)
```
roycaihw: nit: `EXPIRY_SKEW_PREVENTION_DELAY` sounds more clear to me in explaining the purpose of the timedelta. Maybe document these constants and update https://github.com/kubernetes-client/python-base/blob/78472de0c929a57bf3bcff469e3bdb7f09a051c6/config/kube_config_test.py#L36 if we want to change the name.

ffledgling:
Bump, got bit by this today. Any idea what needs to happen to move the needle on this?

PR #75, opened 2018-07-18
Attempt to implement exec-plugins support in kubeconfig

dovreshef:
Hi

This is my attempt to implement exec plugin support. 
I've only tested this using Python3 with EKS.
I'm also only supporting token authentication for now, since it was not clear to me how to support anything else.
I'm willing to do the necessary changes to get it to merge, but I might need help.

Thanks,

### DIFF
--- a/config/exec_provider.py
+++ b/config/exec_provider.py
@@ -1,0 +1,90 @@
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import json
+import os
+import subprocess
+import sys
+
+from .config_exception import ConfigException
+
+
+class ExecProvider(object):
+    """
+    Implementation of the proposal for out-of-tree client authentication providers
+    as described here --
+    https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/kubectl-exec-plugins.md
+
+    Missing from implementation:
+
+    * TLS cert support
+    * caching
+    """
+
+    def __init__(self, exec_config):
+        for key in ['command', 'apiVersion']:
+            if key not in exec_config:
+                raise ConfigException(
+                    'exec: malformed request. missing key \'%s\'' % key)
+        self.api_version = exec_config['apiVersion']
+        self.args = [exec_config['command']]
+        if 'args' in exec_config:
+            self.args.extend(exec_config['args'])
+        self.env = os.environ.copy()
+        if 'env' in exec_config:
+            additional_vars = {}
+            for item in exec_config['env']:
+                name = item['name']
+                value = item['value']
+                additional_vars[name] = value
+            self.env.update(additional_vars)
+
+    def run(self, previous_response=None):
+        kubernetes_exec_info = {
+            'apiVersion': self.api_version,
+            'kind': 'ExecCredential',
+            'spec': {
+                'interactive': sys.stdout.isatty()
+            }
+        }
+        if previous_response:
+            kubernetes_exec_info['spec']['response'] = previous_response
+        self.env['KUBERNETES_EXEC_INFO'] = json.dumps(kubernetes_exec_info)
+        process = subprocess.Popen(
+            self.args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            env=self.env,
+            universal_newlines=True)
+        (stdout, stderr) = process.communicate()
+        exit_code = process.wait()
+        if exit_code != 0:
+            msg = 'exec: process returned %d' % exit_code
+            stderr = stderr.strip()
+            if stderr:
+                msg += '. %s' % stderr
+            raise ConfigException(msg)
+        try:
+            data = json.loads(stdout)
+        except ValueError as de:
+            raise ConfigException(
+                'exec: failed to decode process output: %s' % de)
+        for key in ('apiVersion', 'kind', 'status'):
+            if key not in data:
+                raise ConfigException(
+                    'exec: malformed response. missing key \'%s\'' % key)
+        if data['apiVersion'] != self.api_version:
+            raise ConfigException(
+                'exec: plugin api version %s does not match %s' %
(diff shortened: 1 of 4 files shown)


dovreshef:
CLA signed.

ltamaster:
I tested this PR and worked, any plan to merge it?
Thanks
Luis

cliffpracht:
👍

anneschuth:
I'd love to see this merged too.

yliaog:
thanks for the pr. has this been tested with python 2.7.12?

cliffpracht:
I have tested this with 2.7.15.

config/exec_provider_test.py
```diff
@@ -0,0 +1,140 @@
...
+from .config_exception import ConfigException
+from .exec_provider import ExecProvider
+
+
+class ExecProviderTest(unittest.TestCase):
+
+    def setUp(self):
+        self.input_ok = {
+            'command': 'aws-iam-authenticator token -i dummy',
+            'apiVersion': 'client.authentication.k8s.io/v1alpha1'
```
yliaog: s/v1alpha1/v1beta1/ for all the tests? the feature is v1beta1 is 1.11

config/exec_provider.py
```diff
@@ -0,0 +1,79 @@
...
+                value = item['value']
+                additional_vars[name] = value
+            self.env.update(additional_vars)
+
+    def run(self, previous_response=None):
+        kubernetes_exec_info = {
+            'apiVersion': self.api_version,
+            'kind': 'ExecCredential',
+            'spec': {
+                'interactive': True
```
yliaog: why is it hard coded to True?

config/exec_provider.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+import subprocess
+import sys
+
+from .config_exception import ConfigException
+
+
+class ExecProvider(object):
+    def __init__(self, exec_config):
+        for key in ['command', 'apiVersion']:
```
yliaog: could you list explicitly what is implemented (v1beta1, right?) as listed in the proposal (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/kubectl-exec-plugins.md), or better, what is left out.

AFAICT, TLS cert, caching are not implemented.

what about refreshing? what to do when token expires?

anneschuth:
How far from getting this merged are we? Is there anything I can do to help this go through faster>

yliaog:
I don't think it's far. I put in some comments, as soon as those are addressed, we can merge it.

macat:
@dovreshef Do you need help fixing @yliaog's comments? Thanks for building this BTW.

dovreshef:
@macat I'd be glad for help. I'll not be able to get to this before Monday.

gitfool:
👍

config/exec_provider.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2016 The Kubernetes Authors.
```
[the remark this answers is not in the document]
dovreshef: Changed.

config/exec_provider_test.py
dovreshef: Changed.

config/exec_provider.py
dovreshef: I have changed it to the same as the process stdout (using `sys.stdout.isatty()`), which is what is done here https://github.com/kubernetes/kubernetes/blob/7b6647a418c660f2c87f183f706b297f1cb573ca/staging/src/k8s.io/client-go/plugin/pkg/client/auth/exec/exec.go#L128 (I think).
But I'm still not clear whether or not I should detect interactive stdin as well. Also, if it is interactive, what else am I supposed to do.

dovreshef:
Hi all. I have updated the PR. 

I'm sorry about the delay, real life intervened.

gitfool:
No worries. I’m using a monkey patch with this change to tide me over. 😉

config/exec_provider.py
dovreshef: TLS cert, caching are indeed not implemented.
> TLS cert
I did not implement it.
My reading of the proposal language
```TLS client certificate support is orthogonal to bearer tokens, but something that we should consider supporting in the future.```
 lead me to think it is not implemented in the go client as well.

> caching
I can try to add caching by basing it on the way it is implemented for the other auth methods, taking into account the `expiry` field.
But there is a problem. Right now `aws-iam-authenticator` does not return  an`expiry` field, even though the returned token is only valid for a short period (15 minutes IIRC). 
The RFC on the other hand specifies: 
```Plugins can indicate their credentials explicit expiry using the Expiry field on the returned ExecCredentials object, otherwise credentials will be cached throughout the lifetime of a program.``` 

I can check the return value if it is 401 and in that case re-execute the plugin, which is what the RFC also recommends. But I do not understand the code well enough for that. 
Can you point me where in the code would it be best to hook up such a mechanism?

> what about refreshing? what to do when token expires?
Right now I do not cache the token at all. So since the plugin always executes, the token received is always fresh.

Comment (yliaog) [reply config/exec_provider.py]: since this is a highly desired feature, and it is orthogonal to the other auth methods. I think it is safe to merge with some missing features. people who need those features can implement them on top of this later.

so could you please add some comments about what features are missing, and probably a link 
 to the design (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/kubectl-exec-plugins.md). then I think we can merge it. Thanks.

Comment (dovreshef) [reply config/exec_provider.py]: Added. Hope it's ok.

File: config/kube_config.py [hunk 5]
```diff
@@ -340,6 +342,16 @@ def _refresh_oidc(self, provider):
         provider['config'].value['id-token'] = refresh['id_token']
         provider['config'].value['refresh-token'] = refresh['refresh_token']
 
+    def _load_from_exec_plugin(self):
+        if 'exec' not in self._user:
+            return
+        status = ExecProvider(self._user['exec']).run()
+        if 'token' not in status:
+            raise ConfigException(
+                'Missing token field in exec plugin response')
```
yliaog: i think it's better to catch all the exceptions and return None in case there is any exception, just to give a chance to other auth methods. that is the general pattern used by other _load_...() methods.

dovreshef: Ok. 
Then there is no reason throw any exception at all in exec_provider.py?
In that case how should I inform the user if the exec process return an error, or missing exec config, etc?

yliaog: it's still good to have exceptions from exec_provider.py, then in kube_config.py, when it catches the exception, it could log it, and proceed.

dovreshef: Done. I have used logging.error since it seems to me to be more of an error than a warning. Since it implies an incorrect exec config / result.

yliaog:
thanks for the PR.

/lgtm
/approve

anneschuth:
thanks a lot!

timoreimann:
For posterity: this one fixes kubernetes-client/python#514

PR #77, opened 2018-07-20
Add support for refreshing Azure tokens.

brendandburns:
@roycaihw 

This requires an update to the 'requirements.txt' to import the `adal` library, but it appears that is in a different repo?

@roycaihw is there advice for how to add dependencies?

Thanks!

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -18,7 +18,9 @@ import datetime
 import json
 import os
 import tempfile
+import time
 
+import adal
 import google.auth
 import google.auth.transport.requests
 import oauthlib.oauth2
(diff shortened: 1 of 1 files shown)


roycaihw:
We use [tox](https://github.com/kubernetes-client/python-base/blob/master/run_tox.sh) to test `python-base` change with `python` HEAD. I think you could add the dependency to [python/test-requirements.txt](https://github.com/kubernetes-client/python/blob/master/test-requirements.txt) first to make the CI passes. And after this PR is merged, when we bump the `python-base` submodule in `python` we can sync the actual [requirements.txt](https://github.com/kubernetes-client/python/blob/master/requirements.txt).

brendandburns:
Thanks, filed https://github.com/kubernetes-client/python/pull/579 for requirements.txt

config/kube_config.py
```diff
@@ -202,10 +204,29 @@ def _load_azure_token(self, provider):
...
+        tenant = config['tenant-id']
+        authority = 'https://login.microsoftonline.com/{}'.format(tenant)
+        context = adal.AuthenticationContext(
+            authority, validate_authority=True,
+        )
+        refresh_token = config['refresh-token']
+        client_id = config['client-id']
+        token_response = context.acquire_token_with_refresh_token(
+            refresh_token, client_id, '00000002-0000-0000-c000-000000000000')
+
```
roycaihw: the CI pycodestyle complains about blank line containing whitespace

PR #78, opened 2018-07-21
The websockets client should not assume everything is UTF-8 encoded on Python 3

ceridwen:
Currently, the Kubernetes websockets client assumes that everything that it receives is encoded in UTF-8.  If you try to return binary data or data using another encoding from a command run inside a container using `connect_get_namespaced_pod_exec`, it will usually raise an exception, though it might also sometimes return mangled output.  This pull request makes it so that `exec`ed commands that return binary will be handled correctly on Python 3.  It also changes the string type of the returned STDOUT and STDERR channels to bytes.  This is consistent with both the websocket-client library and the stdlib subprocess module, both of which return bytes on Python 3.  Clients using this library can then do their own decoding, since they presumably know which encoding they expect to receive from the application they're `exec`ing.

### DIFF
--- a/stream/stream.py
+++ b/stream/stream.py
@@ -26,9 +26,15 @@ def stream(func, *args, **kwargs):
 
         return ws_client.websocket_call(config, *args, **kwargs)
 
+    def _intercept_deserialize_call(response, response_type):
+        return response.data
+
     prev_request = func.__self__.api_client.request
+    prev_deserialize = func.__self__.api_client.deserialize
     try:
         func.__self__.api_client.request = _intercept_request_call
+        func.__self__.api_client.deserialize = _intercept_deserialize_call
         return func(*args, **kwargs)
     finally:
         func.__self__.api_client.request = prev_request
+        func.__self__.api_client.deserialize = prev_deserialize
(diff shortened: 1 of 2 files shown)


ceridwen:
The tests are failing because of my proposed API change, as they depend on the websocket call returning JSON and `deserialize` turning it into Python objects.  This is what I want to discuss.

tbarrella:
Hello, ran into an issue due to this as well, any thoughts on the status of this? I guess API changes are a concern; potentially could add something like an `encoding="utf-8"` param to avoid breaking changes (`None` -> bytes, as in `subprocess` functions), or make a new function that returns bytes?

elemental-lf:
I've just written code which now depends on the fact that the decoding is done by the websockets client, so I'd prefer a non-API breaking change. But I also see the need to return binary data under some circumstances. I like @tbarrella's idea to add an optional `encoding` parameter which can be set to `None` to return binary data but is set to `utf-8` by default.

scottilee:
Hey @ceridwen, can you fix the conflict? I can get someone to review it afterwards.

Since this PR has been open for a while, if it is no longer valid please close it.

PR #79, opened 2018-07-30
Fix base64 padding for kube config

### LINKED ISSUE #65: OIDC auth uses incorrect base64 decoding
We had a customer report the following error with this client:

```
File ".../k8s_client.py", line 6, in <module> 
config.load_kube_config() 
File ".../venv/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 473, in load_kube_config 
loader.load_and_set(config) 
File ".../venv/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 354, in load_and_set 
self._load_authentication() 
File ".../venv/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 185, in _load_authentication 
if self._load_oid_token(): 
File ".../venv/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 236, in _load_oid_token 
base64.b64decode(parts[1]).decode('utf-8') 
File ".../python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/base64.py", line 87, in b64decode 
return binascii.a2b_base64(s) 
binascii.Error: Incorrect padding
```

Which appears to originate from here: https://github.com/kubernetes-client/python-base/blob/789de6a60dea9442ee47601a788d92c653ea69e9/config/kube_config.py#L236

```
jwt_attributes = json.loads(
    base64.b64decode(parts[1]).decode('utf-8')
)
```

JWTs aren't encoded using standard base64 encoding, they use URL encoding without the final padding

```
   Base64url Encoding
      Base64 encoding using the URL- and filename-safe character set
      defined in Section 5 of RFC 4648 [RFC4648], with all trailing '='
      characters omitted (as permitted by Section 3.2) and without the
      inclusion of any line breaks, whitespace, or other additional
      characters.  Note that the base64url encoding of the empty octet
      sequence is the empty string.  (See Appendix C for notes on
      implementing base64url encoding without padding.)
```

https://tools.ietf.org/html/rfc7515#section-2

So "hello world" should become `aGVsbG8gd29ybGQ`, not `aGVsbG8gd29ybGQ=` https://play.golang.org/p/vFrVzr9uyAQ

Python's default base64 library doesn't handle this encoding and spits out the same exception our customer's seeing:

```
$ python3 -c 'import base64; base64.b64decode("aGVsbG8gd29ybGQ")'
  File "<string>", line 1, in <module>
  File "/usr/lib64/python3.6/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Incorrect padding
```

bpicolo:
Fixes #65 (pulled in / updated another open PR to get this fix moving + tweaked tests)

JWT spec requires tokens be de-padded base64 (so that they're URLsafe), but they need to be re-padded to properly decode.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -257,18 +257,18 @@ class KubeConfigLoader(object):
 
         if any(char in token for char in reserved_characters):
             # Invalid jwt, as it contains url-unsafe chars
-            return None
+            return
 
         parts = token.split('.')
         if len(parts) != 3:  # Not a valid JWT
-            return None
+            return
 
         padding = (4 - len(parts[1]) % 4) * '='
         if len(padding) == 3:
             # According to spec, 3 padding characters cannot occur
             # in a valid jwt
             # https://tools.ietf.org/html/rfc7515#appendix-C
-            return None
+            return
 
         if PY3:
             jwt_attributes = json.loads(
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -252,12 +252,23 @@ class KubeConfigLoader(object):
         if 'config' not in provider:
             return
 
-        parts = provider['config']['id-token'].split('.')
+        reserved_characters = frozenset(["=", "+", "/"])
+        token = provider['config']['id-token']
 
+        if any(char in token for char in reserved_characters):
+            # Invalid jwt, as it contains url-unsafe chars
+            return None
+
+        parts = token.split('.')
         if len(parts) != 3:  # Not a valid JWT
             return None
 
         padding = (4 - len(parts[1]) % 4) * '='
+        if len(padding) == 3:
+            # According to spec, 3 padding characters cannot occur
+            # in a valid jwt
+            # https://tools.ietf.org/html/rfc7515#appendix-C
+            return None
 
         if PY3:
             jwt_attributes = json.loads(
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -257,13 +257,15 @@ class KubeConfigLoader(object):
         if len(parts) != 3:  # Not a valid JWT
             return None
 
+        padding = (4 - len(parts[1]) % 4) * '='
+
         if PY3:
             jwt_attributes = json.loads(
-                base64.b64decode(parts[1]).decode('utf-8')
+                base64.b64decode(parts[1] + padding).decode('utf-8')
             )
         else:
             jwt_attributes = json.loads(
-                base64.b64decode(parts[1] + "==")
+                base64.b64decode(parts[1] + padding)
             )
 
         expire = jwt_attributes.get('exp')
--- a/config/kube_config_test.py
+++ b/config/kube_config_test.py
@@ -43,8 +43,8 @@ def _base64(string):
     return base64.encodestring(string.encode()).decode()
 
 
-def _unpadded_base64(string):
-    return base64.b64encode(string.encode()).decode().rstrip('=')
+def _urlsafe_unpadded_b64encode(string):
+    return base64.urlsafe_b64encode(string.encode()).decode().rstrip('=')
 
 
 def _format_expiry_datetime(dt):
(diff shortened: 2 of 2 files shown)


bpicolo:
@k8s-ci-robot should be CLAd now

zybjcdl:
@bpicolo When could we get this fix? We need to run with python 3.6.5

config/kube_config.py
```diff
@@ -257,13 +257,15 @@ def _load_oid_token(self, provider):
         if len(parts) != 3:  # Not a valid JWT
             return None
 
+        padding = (4 - len(parts[1]) % 4) * '='
+
         if PY3:
             jwt_attributes = json.loads(
-                base64.b64decode(parts[1]).decode('utf-8')
+                base64.b64decode(parts[1] + padding).decode('utf-8')
```
roycaihw: JWT uses `base64.urlsafe_b64decode` instead of `base64.b64decode` (and for encoding as well), as `+` and `/` are not url safe

config/kube_config_test.py
```diff
@@ -87,11 +91,13 @@ def _raise_exception(st):
 
 TEST_OIDC_TOKEN = "test-oidc-token"
 TEST_OIDC_INFO = "{\"name\": \"test\"}"
-TEST_OIDC_BASE = _base64(TEST_OIDC_TOKEN) + "." + _base64(TEST_OIDC_INFO)
+TEST_OIDC_BASE = _unpadded_base64(
+    TEST_OIDC_TOKEN) + "." + _unpadded_base64(TEST_OIDC_INFO)
 TEST_OIDC_LOGIN = TEST_OIDC_BASE + "." + TEST_CLIENT_CERT_BASE64
```
roycaihw: The signature (`TEST_CLIENT_CERT_BASE64`) should also be url safe (base64url encoded and padding omitted).

config/kube_config.py
```diff
@@ -257,13 +257,15 @@ def _load_oid_token(self, provider):
         if len(parts) != 3:  # Not a valid JWT
```
roycaihw: Could you add a check to make sure the JWT doesn't contain url reserved characters (`=`, `+` and `/`)?

roycaihw:
After more reading, I'm thinking if the previous [padding implementation](https://github.com/kubernetes-client/python-base/blob/7d1e4495a8222f9992c1a0d4bd0cc96832b82ec2/config/kube_config.py#L273) (padding with `"=="`) was intended.

According to [base64 encoding](https://tools.ietf.org/html/rfc4648#section-4) and [JWS](https://tools.ietf.org/html/rfc7515#appendix-C) doc, base64 (also apply to base64url) padding can only have three cases: 
 - no padding
 - two "=" padding characters
 - one "=" padding character

Having three "=" padding characters is illegal. 

According to https://gist.github.com/perrygeo/ee7c65bb1541ff6ac770 and [python binascii_a2b_base64 implementation](https://github.com/python/cpython/blob/2.7/Modules/binascii.c#L403-L419), extra paddings after previous quad are ignored. Therefore for a valid base64-encoded string with padding stripped, adding "==" would guarantee it can be safely decoded, and for an illegal string (e.g. "A===" being "A" when stripped) adding "==" would raise error.

In Python 3.7 the implementation distinguishes illegal string ("A==") from incorrect padding ("AQ="): https://github.com/python/cpython/blob/3.7/Modules/binascii.c#L513-L521. Prior 3.7 the error messages are the same "Incorrect padding". 

https://github.com/kubernetes-client/python-base/issues/65 had this issue because we didn't implement padding for python 3. The padding implementation for python 2 was correct (probably intended?) but it didn't distinguish illegal string from incorrect padding. I suggest we could add illegal base64url string check like https://tools.ietf.org/html/rfc7515#appendix-C

cc @ltamaster who implemented oidc auth
cc @hanikesn @joppa27 who also contributed on this fix

bpicolo:
@roycaihw  fair - I didn't notice the addendum in the JWS RFC. What's the optimal path forward?

roycaihw:
@bpicolo I think your fix is fine for both Python 2 and 3. I'd suggest that we can add a check and raise error if `len(padding) == 3`, so that the versions prior 3.7 can better distinguish malformed token and incorrect padding. 

Please also fix the decoding and testcases around oidc id-token (JWT) to use [base64url-encoding/decoding](https://tools.ietf.org/html/rfc7519#section-3).

bpicolo:
@roycaihw Had the first chance/attention span in, clearly quite a while, to address the things we talked about long ago, because I get emails about this on occasion from people referencing it, heh.

The actionables I tweaked:
- Fail if padding length is 3
- Fail if reserved characters are present
- Urlsafe test b6eencode

Hopefully this is good to go, though a potentially useful follow up might be noisier auth errors

micw523:
Hi @bpicolo - 
It looks like that your tests are somewhat incomplete and decreased the code coverage for the bot. It is possible for you to add some tests that you mentioned, e.g. if padding length is 3, etc?

bpicolo:
@micw523  Done.

Where is code coverage report now? Last one I see is in July?

micw523:
I think the bot refreshes its comment every once a while - it should refresh itself.

Otherwise, the CI says your coverage is 92%.

bpicolo:
@micw523  Ahhh I see that now, gotcha

micw523:
Are you running into import problems with adal? I think part of the reason might be that you opened this PR a long time ago and the python repo has ditched adal as an absolute requirement. It was a recent change. Fast forwarding to the master branch might help?

I agree on the difficulty on running the tests. The documentation for this library is quite sparse in its current state.

config/kube_config.py
```diff
@@ -252,18 +252,31 @@ def _load_oid_token(self, provider):
         if 'config' not in provider:
             return
 
-        parts = provider['config']['id-token'].split('.')
+        reserved_characters = frozenset(["=", "+", "/"])
+        token = provider['config']['id-token']
 
+        if any(char in token for char in reserved_characters):
+            # Invalid jwt, as it contains url-unsafe chars
+            return None
```
micw523: It looks like that in the original code a single ```return``` is used. Is it possible if we make the return statements consistent?

config/kube_config.py
micw523: It might be still worthwhile to do safe decode here?

bpicolo:
@micw523  cleaned up the return. re: safe decode (can't reply to that comment), we're breaking above if there are url-unsafe characters, so seems unnecessary. 

In theory, this is ready

stevenmanton:
Just pinging this thread to see if there are updates. This bug still blocks our development without manually editing the offending line. Is this PR just awaiting lgtm?

roycaihw:
Sorry for the delay in review. I will do another pass this week

carlescere:
Bump

roycaihw:
my apologies.. I was busy with k/k 1.15 freeze. It looks like my comments were addressed a long time ago :)

> re: safe decode (can't reply to that comment), we're breaking above if there are url-unsafe characters, so it should be safe as is.

Fair enough. I re-ran the travis CI and the [`test_create_apiservice_from_yaml_with_conflict`](https://travis-ci.org/kubernetes-client/python-base/jobs/513453059) failure looks unrelated. I tend to merge this. Thanks for the fix!

/lgtm
/approve

rcurrie:
Just installed released 10.0.0 fresh from pip and tested against the [Pacific Research Platform](http://pacificresearchplatform.org/) k8s cluster using python3 and it works now! Thank You!

Issue #84, opened 2018-09-14
Buggy conditional in kube_config.py (TypeError: '<' not supported between instances of 'int' and 'time.struct_time')


wunzeco:
https://github.com/kubernetes-client/python-base/blob/d68e456190ec9dc8bbac5fd922f363b3b90b7a90/config/kube_config.py#L211

The line above fails with the error:
```
 File "/Users/ogonna/virtualenvs/k8stlsingress-Q6OI3mKm/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 505, in load_kube_config
    loader.load_and_set(config)
  File "/Users/ogonna/virtualenvs/k8stlsingress-Q6OI3mKm/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 386, in load_and_set
    self._load_authentication()
  File "/Users/ogonna/virtualenvs/k8stlsingress-Q6OI3mKm/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 183, in _load_authentication
    if self._load_auth_provider_token():
  File "/Users/ogonna/virtualenvs/k8stlsingress-Q6OI3mKm/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 198, in _load_auth_provider_token
    return self._load_azure_token(provider)
  File "/Users/ogonna/virtualenvs/k8stlsingress-Q6OI3mKm/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 209, in _load_azure_token
    if int(provider['config']['expires-on']) < time.gmtime():
TypeError: '<' not supported between instances of 'int' and 'time.struct_time'
```

Suggested fix:
```
            if time.gmtime(int(provider['config']['expires-on'])) < time.gmtime():
                self._refresh_azure_token(provider['config'])
```

sdktr:
This fix (and the original behavior as well) causes issues for me. My assumption is the date format when the kubeconfig is accessed by Windows as well, since I'm running on Windows Subsystem for Linux and use a mounted file as Kubeconfig?

```File "/usr/local/lib/python2.7/dist-packages/kubernetes/config/kube_config.py", line 209, in _load_azure_token
        if time.gmtime(int(provider['config']['expires-on'])) < time.gmtime():
    ValueError: invalid literal for int() with base 10: '2018-10-18 00:52:29.044727'
```

mathieuherbert:
Hi, 
[...]

PR #85, opened 2018-09-14
Fix conditional checking an int against time.struct_time

wunzeco:
Fixes this issue https://github.com/kubernetes-client/python-base/issues/84

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -208,7 +208,7 @@ class KubeConfigLoader(object):
         if 'access-token' not in provider['config']:
             return
         if 'expires-on' in provider['config']:
-            if int(provider['config']['expires-on']) < time.gmtime():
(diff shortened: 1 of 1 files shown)


config/kube_config.py
```diff
@@ -208,7 +208,7 @@ def _load_azure_token(self, provider):
         if 'access-token' not in provider['config']:
             return
         if 'expires-on' in provider['config']:
-            if int(provider['config']['expires-on']) < time.gmtime():
+            if time.gmtime(int(provider['config']['expires-on'])) < time.gmtime():
```
yliaog: what about if int(provider['config']['expires-on'])) < time.time()

yliaog: also could you add a test case for it?

sysincz: Tested on azure. working for me thanks

PR #86, opened 2018-09-17
fix: read config data with bytes (python3)

tomplus:
I came across a problem where `kubeconfig` has certificates stored as binary data.

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: !!binary |
      TAMwdExTSUNSVWRKVGlCRFJWOlVTVVpKUTBLVVJTMHRMUzB0A2sxSlNVWm1SRU5EUVRKVFowRjNT
      ............
      bZBKUkhZeFIzQnNARGRuUFQwS0xTMHFMUzFGVGtRZ1EwVlNWRWWHU1VOQlZFZXRMUzB1TFE9PQ==
...
```

Python 3 sees this as bytes and the encoding problem appears.

```
  File "examples/example1.py", line 33, in <module>
    main()
  File "examples/example1.py", line 22, in main
    config.load_kube_config()
  File "/tmp/python/kubernetes/config/kube_config.py", line 520, in load_kube_config
    loader.load_and_set(config)
  File "/tmp/python/kubernetes/config/kube_config.py", line 402, in load_and_set
    self._load_cluster_info()
  File "/tmp/python/kubernetes/config/kube_config.py", line 381, in _load_cluster_info
    file_base_path=self._config_base_path).as_file()
  File "/tmp/python/kubernetes/config/kube_config.py", line 104, in as_file
    base64.decodestring(self._data.encode()))
AttributeError: 'bytes' object has no attribute 'encode'
```

In this PR I've added a check if `_data` are the `str` type.

Thanks.


config/kube_config_test.py
```diff
@@ -65,6 +65,9 @@ def _raise_exception(st):
 TEST_DATA = "test-data"
 TEST_DATA_BASE64 = _base64(TEST_DATA)
 
+TEST_DATA_BYTES = b"test-data"
+TEST_DATA_BASE64_BYTES = b"dGVzdC1kYXRh"
```
roycaihw: nit: you can use `TEST_DATA.encode()` or `bytes(TEST_DATA, encoding='utf-8')` instead of introducing new variables

tomplus: +1, fixed

roycaihw: Thanks! Could you squash the commits?

tomplus: Done

tomplus:
I have no idea how to test it again. The previous build failed but it looks like an Travis' issue.

PR #91, opened 2018-10-05
ExecProvider errors when parsing null values

fillbit:
When executing a command like kubectl config use-context <ctx> kubectl rewrite the the kubeconfig file and adds null to optional parameters. For example a a config with 
```
users:
- name: Name1
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      command: heptio-authenticator-aws
```

is rewritten as:

```
users:
- name: Name1
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args: null
      command: heptio-authenticator-aws
      env: null
```

When the ExecParser reads this it returns the error: 
```
ERROR: Invalid kube-config file. Expected key args in kube-config/users[name=Name1]/user/exec
```

This PR adds a test case and a fix for this issue.

### DIFF
--- a/config/exec_provider.py
+++ b/config/exec_provider.py
@@ -32,16 +32,21 @@ class ExecProvider(object):
     """
 
     def __init__(self, exec_config):
+        """
+        exec_config must be of type ConfigNode because we depend on
+        safe_get(self, key) to correctly handle optional exec provider
+        config parameters.
+        """
         for key in ['command', 'apiVersion']:
             if key not in exec_config:
                 raise ConfigException(
                     'exec: malformed request. missing key \'%s\'' % key)
         self.api_version = exec_config['apiVersion']
         self.args = [exec_config['command']]
-        if 'args' in exec_config:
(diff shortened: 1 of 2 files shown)


fillbit:
Signed CLA?

fillbit:
signed CLA?

fillbit:
I've updated the PR to use safe_get and the unit tests to use ConfigNode instead of dict() so they work correctly. I also updated the input_ok to include a `env: null` value and removed my test case as it is now redundant

roycaihw:
@fillbit Thanks. Could you also add a comment in `ExecProvider` documenting that `__init__` expects `exec_config` to be of `ConfigNode` class, which implements the `safe_get(self, key)` method? Otherwise LGTM

roycaihw:
Please fix the pycodestyle error for travis CI to pass, and squash commits before submit :)

fillbit:
@roycaihw @mbohlool  Are there any other issue that need to be addressed?

PR #92, opened 2018-10-06
Refresh GCP tokens on retrieval by overriding client config method.

### LINKED ISSUE #59: GCP Token is not refreshed

TrevorEdwards:
For GCP auth, this overrides the method used by the swagger-generated kubernetes client to get an authorization token, allowing us to use existing python-base code to refresh the GCP auth token. This should [fix #59], as API calls that happen > 1 hour after the client is populated will have the token refreshed around 5 minutes before expiration.

For reference, https://github.com/swagger-api/swagger-codegen/blob/4607a90d7b69463a0ae8fc94fac68fc95d80965e/modules/swagger-codegen/src/main/resources/python/configuration.mustache#L215 is the swagger generator and https://github.com/kubernetes-client/python/blob/master/kubernetes/client/configuration.py#L195 is today's generated client.

I also considered using a custom dict (https://github.com/TrevorEdwards/python-base/commit/d5c8cdc696805bedb4bfd4a8eb8746631c22a276), but that solution seemed too hacky.


config/kube_config_test.py
```diff
@@ -509,7 +521,22 @@ class TestKubeConfigLoader(BaseTestCase):
...
+                            "expiry": TEST_TOKEN_EXPIRY_PAST,  # always in past
+                        }
+                    },
+                    "token": TEST_DATA_BASE64,  # should be ignored
+                    "username": TEST_USERNAME,  # should be ignored
+                    "password": TEST_PASSWORD,  # should be ignored
+                }
+            },
+            {
+                "name": "expired_gcp_refresh",
```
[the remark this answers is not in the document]
TrevorEdwards: Yes. Added a comment to note that.

config/kube_config.py
```diff
@@ -392,7 +392,21 @@ def _load_cluster_info(self):
...
+        if self._using_gcp_auth_provider():
+            # GCP auth tokens must be refreshed regularly, but swagger expects
+            # a constant token. Replace the swagger-generated client config's
+            # get_api_key_with_prefix method with our own to allow automatic
+            # token refresh.
+            def _gcp_get_api_key(*args):
+                return self._load_gcp_token(self._user['auth-provider'])
+            client_configuration.get_api_key_with_prefix = _gcp_get_api_key
         if 'token' in self.__dict__:
             client_configuration.api_key['authorization'] = self.token
```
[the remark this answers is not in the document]
TrevorEdwards: I added comments to make that behavior more clear, unless you mean that you want this documented somewhere else.

The only existing workaround I know of will catch exceptions when an API call fails (via a wrapper) and then reload the config. I believe with this change, they will never hit that exception.

config/kube_config_test.py
```diff
@@ -630,16 +659,20 @@ def test_load_user_token(self):
         self.assertEqual(BEARER_TOKEN_FORMAT % TEST_DATA_BASE64, loader.token)
 
     def test_gcp_no_refresh(self):
-        expected = FakeConfig(
-            host=TEST_HOST,
-            token=BEARER_TOKEN_FORMAT % TEST_DATA_BASE64)
-        actual = FakeConfig()
+        fake_config = FakeConfig()
+        # swagger-generated config has this, but FakeConfig does not.
+        self.assertFalse(hasattr(fake_config, 'get_api_key_with_prefix'))
```
mbohlool: can we add a test to make sure swagger-generated config has this?

config/kube_config.py
```diff
@@ -392,8 +392,24 @@ def _load_cluster_info(self):
...
+
     def _set_config(self, client_configuration):
+        if self._using_gcp_auth_provider():
+            # GCP auth tokens must be refreshed regularly, but swagger expects
+            # a constant token. Replace the swagger-generated client config's
+            # get_api_key_with_prefix method with our own to allow automatic
+            # token refresh.
+            def _gcp_get_api_key(*args):
+                return self._load_gcp_token(self._user['auth-provider'])
+            client_configuration.get_api_key_with_prefix = _gcp_get_api_key
```
mbohlool: I am ok with this hack. just to make sure we are protected when things change in configuration class of generated client, can we add proper tests to make sure get_api_key_with_prefix exists and being called the way we expected it.
I can't tell from tests that we are protected against changes in configuration class. please add them if they are missing. e.g. if get_api_key_with_prefix got rename, or if it is being called but it returns something else. I see a whitebox test for get_api_key_with_prefix, but it won't protect us against get_api_key_with_prefix being removed or renamed in configuration class).

config/kube_config_test.py
TrevorEdwards: Done.

config/kube_config.py
TrevorEdwards: Done. I added three new test cases which I believe cover all of the concerns that you list.

TrevorEdwards:
Build is failing, but I think it's due to the changes introduced in https://github.com/kubernetes-client/python/commit/e63905302fd4da45471253876df2c4fb45781fbf#diff-07d29d9544bd4183a0a05fa23257e92d. I'll look into it.

micw523:
Hi @TrevorEdwards, I introduced this code change. It is likely due to an out-of-date ubuntu version change when we switched to a higher version of k8s on the server. Fixing it right now #99

TrevorEdwards:
Build's fixed, thanks @micw523

PR #93, opened 2018-10-16
Making watch work with read_namespaced_pod_log

mitar:
Fixes https://github.com/kubernetes-client/python/issues/199.

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -151,7 +151,7 @@ class Watch(object):
                 if self.resource_version is not None:
                     kwargs['resource_version'] = self.resource_version
                 else:
-                    break
+                    self._stop = True
 
             if timeouts or self._stop:
                 break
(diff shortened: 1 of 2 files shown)


yliaog:
please add a test case for this.

watch/watch.py
```diff
@@ -132,9 +142,12 @@ def stream(self, func, *args, **kwargs):
                     if self._stop:
                         break
             finally:
-                kwargs['resource_version'] = self.resource_version
                 resp.close()
                 resp.release_conn()
+                if self.resource_version is not None:
+                    kwargs['resource_version'] = self.resource_version
+                else:
```
micw523: Do you need the else: break here?

mitar: Yes. The log streaming does not have `resource_version`. So once the response finishes, you have to stop.

mitar:
@yliaog I have added tests.

@micw523 Please review and maybe merge this?

micw523:
As much as I want to merge this I don't have ownership of this repo :)
/lgtm
/cc @roycaihw

watch/watch.py
mitar: In fact that was not good because using break inside finally swallows up the exception. I change it to setting a flag.

Issue #62, opened 2018-04-24
load_kube_config fails if KUBECONFIG contains more than one file


joar:
KUBECONFIG is a list of colon-deliminated config files.
https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#the-kubeconfig-environment-variable.

load_kube_config without arguments throws:
```pytb
FileNotFoundError: [Errno 2] No such file or directory: '/home/joar/git/foo/.kube-config:/home/joar/.kube/config'
```

matohavo:
Bump. Still an issue

aseempatni:
I'm also facing the same issue when KUBECONFIG contains multiple files.

tomplus:
@aseempatni which version of the library do you have? It should work with the latest one.

/remove-lifecycle stale

aseempatni:
Ah, I see. I'm using 8.0.1. It was fixed after that I assume? Lemme give a shot with the latest one.

tomplus:
Yes, v. 10.0.0a1 should work.

PR #94, opened 2018-10-19
feat: merging kubeconfig files

tomplus:
I'm trying to add support for multiple kubeconfig file. I prepared an additional class which merges configs and produces ConfigNode which can be passed directly to KubeConfigLoader. KubeConfigLoader and other public methods are backward compatible. I had to extend the ConfigNode by adding path to source kubeconfig too, because paths to certificates are relative.

Please let me know what you think.

Related issue: https://github.com/kubernetes-client/python-base/issues/62

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -14,10 +14,12 @@
 
 import atexit
 import base64
+import copy
 import datetime
 import json
 import logging
 import os
+import platform
 import tempfile
 import time
 
@@ -38,6 +40,7 @@ from .dateutil import UTC, format_rfc3339, parse_rfc3339
 
 EXPIRY_SKEW_PREVENTION_DELAY = datetime.timedelta(minutes=5)
 KUBE_CONFIG_DEFAULT_LOCATION = os.environ.get('KUBECONFIG', '~/.kube/config')
+ENV_KUBECONFIG_PATH_SEPARATOR = ';' if platform.system() == 'Windows' else ':'
 _temp_files = {}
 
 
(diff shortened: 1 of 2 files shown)


tomplus:
@roycaihw could you take a look?

tomplus:
@roycaihw Have you had time to look into this?

config/kube_config.py
```diff
@@ -482,26 +500,100 @@ def get_with_name(self, name, safe=False):
...
+    return KubeConfigLoader(
+        config_dict=kcfg.config,
+        config_base_path=None,
+        **kwargs)
 
 
 def list_kube_config_contexts(config_file=None):
 
     if config_file is None:
-        config_file = os.path.expanduser(KUBE_CONFIG_DEFAULT_LOCATION)
```
micw523: I think ```os.path.expanduser``` should be used since there are Windows users.

config/kube_config_test.py
```diff
@@ -987,5 +988,167 @@ def fake_get_api_key_with_prefix(identifier):
...
+                "user": {
+                    "token": TEST_DATA_BASE64,
+                    "username": TEST_USERNAME,  # should be ignored
+                    "password": TEST_PASSWORD,  # should be ignored
+                }
+            },
+        ]
+    }
+
+    def _create_mulit_config(self):
```
micw523:
```suggestion
    def _create_multi_config(self):
```

config/kube_config.py
```diff
@@ -523,18 +615,12 @@ def load_kube_config(config_file=None, context=None,
...
-
-    config_persister = None
-    if persist_config:
-        def _save_kube_config(config_map):
-            with open(config_file, 'w') as f:
-                yaml.safe_dump(config_map, f, default_flow_style=False)
-        config_persister = _save_kube_config
+        config_file = KUBE_CONFIG_DEFAULT_LOCATION
```
micw523: Same ^

config/kube_config.py
roycaihw: handled in [KubeConfigMerger](https://github.com/kubernetes-client/python-base/pull/94/files#diff-a6c16d793c23d6a28f9255727fcd3373R535)

config/kube_config_test.py
tomplus: Fixed, thanks for spotting it.

tomplus:
@roycaihw Thank you!

PR #101, opened 2018-11-12
[WIP] Remove absolute dependence of adal

micw523:
Fixes kubernetes-client/python#668
Fixes kubernetes-client/python#645
Will need to be updated concurrently with a PR from the main repo.
Task list:
[x] Use try / except to handle adal import
[ ] modify setup file to add the adal option [main repo]
[ ] add adal option to the install section of documentation [main repo]
[ ] change the main test interface to install k8s with adal
[ ] add a test env and a minimal test that does not include adal

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -21,7 +21,11 @@ import os
 import tempfile
 import time
 
-import adal
+try:
+    import adal
+except ImportError:
+    pass
+
 import google.auth
 import google.auth.transport.requests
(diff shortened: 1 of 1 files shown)


dims:
@micw523 @iamneha since this is a WIP and has a task list ... 

- @micw523 can you please turn the task list into issues?
- Also let's close this and leave 108 as we need to pick one or the other :)

Sounds good?

PR #102, opened 2018-11-15
Handle error events in watch

### LINKED ISSUE #57: Watch stream should handle HTTP error before unmarshaling event
I could be mistaken but looking at the infinite loop for the watch stream doesn't handle the case when you receive an event that is expired i.e. a HTTP status code of 410. 

``` python
        while True:
            resp = func(*args, **kwargs)
            try:
                    yield self.unmarshal_event(line, return_type)
                    if self._stop:
                        break
            finally:
                kwargs['resource_version'] = self.resource_version
                resp.close()
                resp.release_conn()
```
Looking at the code it seems that if the event is expired then resp should return something along the lines of 
```
{'raw_object': {u'status': u'Failure', u'kind': u'Status', u'code': 410, u'apiVersion': u'v1', u'reason': u'Gone', u'message': u'too old resource version: 2428 (88826)', u'metadata': {}}, u'object': {'api_version': 'v1',
 'kind': 'Status',
```
And `unmarshall_event` should fail to deserialize the object and break. And `self.resource_version` should just be stuck on the resource_version of the event that was expired. 

Am I missing something here?

fabxc:
Raise an ApiException for error events that indicate a watch failure
despite the HTTP response indicating success.

Fixes #57

The ApiException constructor indicates that it's intended to be used manually as well. Thus it seemed convenient to use and is the least likely to break any callers that likely have handling for that exception in place already.

@roycaihw

### DIFF
--- a/watch/watch_test.py
+++ b/watch/watch_test.py
@@ -16,6 +16,8 @@ import unittest
 
 from mock import Mock
 
+from kubernetes import client
+
 from .watch import Watch
 
 
@@ -166,6 +168,32 @@ class WatchTests(unittest.TestCase):
         except KeyError:
             pass
             # expected
+
+        fake_api.get_thing.assert_called_once_with(
+            _preload_content=False, watch=True)
+        fake_resp.read_chunked.assert_called_once_with(decode_content=False)
+        fake_resp.close.assert_called_once()
+        fake_resp.release_conn.assert_called_once()
+
+    def test_watch_with_error_event(self):
+        print("start")
+        fake_resp = Mock()
+        fake_resp.close = Mock()
+        fake_resp.release_conn = Mock()
+        fake_resp.read_chunked = Mock(
+            return_value=[
+                '{"type": "ERROR", "object": {"code": 410, '
+                '"reason": "Gone", "message": "error message"}}\n'])
+
+        fake_api = Mock()
+        fake_api.get_thing = Mock(return_value=fake_resp)
+
+        w = Watch()
+        try:
+            for _ in w.stream(fake_api.get_thing):
+                self.fail(self, "Should fail with ApiException.")
+        except client.rest.ApiException:
+            pass
 
         fake_api.get_thing.assert_called_once_with(
             _preload_content=False, watch=True)
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -128,7 +128,14 @@ class Watch(object):
             resp = func(*args, **kwargs)
             try:
                 for line in iter_resp_lines(resp):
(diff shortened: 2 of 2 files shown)


watch/watch_test.py
```diff
@@ -15,10 +15,10 @@
 import unittest
 
 from mock import Mock
+from kubernetes import client
 
 from .watch import Watch
 
-
 class WatchTests(unittest.TestCase):
```
micw523: Please run your code through autopep8.

mitar:
Should we really throw an exception here and not just retry the watch?

mitar:
On a better look, I do not think this fully addresses #57 in any case. To me `unmarshal_event` already fails because it cannot really convert the response data to API response of a rellevant type. I get:

```
...
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/watch/watch.py", line 141, in stream
    yield self.unmarshal_event(line, return_type)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/watch/watch.py", line 94, in unmarshal_event
    js['object'] = self._api_client.deserialize(obj, return_type)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/api_client.py", line 236, in deserialize
    return self.__deserialize(data, response_type)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/api_client.py", line 276, in __deserialize
    return self.__deserialize_model(data, klass)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/api_client.py", line 622, in __deserialize_model
    instance = klass(**kwargs)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/models/v1_event.py", line 107, in __init__
    self.involved_object = involved_object
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/models/v1_event.py", line 266, in involved_object
    raise ValueError("Invalid value for `involved_object`, must not be `None`")
ValueError: Invalid value for `involved_object`, must not be `None`
```

So, I think: this should be handled sooner, and it should just automatically resume if possible.

mitar:
I made slightly improved (in my view) version here: https://github.com/kubernetes-client/python-base/compare/master...mitar:retry-watch

Would you maybe be interested in incorporating those changes here and updating tests here to trigger those other conditions as well? For example, I am using `list_event_for_all_namespaces` and I am getting the error above with just your patch. I had to add a check to not try to unmarshall it. We should have a test to make sure without that check code fails. I do not understand why in your case this does not happen, but maybe you used a different resource type which happens to be able to be unmarshalled even if it is in fact an error.

PR #103, opened 2018-11-20
Utility for adding nonblocking interface for blocking io object.

zq-david-wang:
Live streaming k8s pod logs is a feature useful for client. But current log api return a blocking io object to client, it is hard for client to proactively close the connection. This utility would help convert the blocking io object into a non-blocking one, follow is a usage example:

```
        v1 = kubernetes.client.CoreV1Api()
        log_stream = poll.Poll(v1.read_namespaced_pod_log(
            name="hello", namespace="default",
            _preload_content=False, follow=True,
            tail_lines=200
        ))
        while True:
            if client_close or log_steam.isclosed():
                break
            c = log_stream.read_until(timeout=0)
            # process the log
            time.sleep(.1)
        log_stream.close()
```

### DIFF
--- a/poll/poll.py
+++ b/poll/poll.py
@@ -1,0 +1,53 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


zq-david-wang:
CLA? WTF?

zq-david-wang:
Is this CLA next step of open source projects........

Issue #88, opened 2018-09-26
stream assumes all data as UTF-8. That's not always the case.


Ark-kun:
```
resp = kubernetes.stream.stream(api.connect_get_namespaced_pod_exec, 'sleeping-bash', namespace='default', command=['bash', '-c', 'echo -n -e \\\\xD4'], stderr=True, stdin=False, stdout=True, tty=False)
```

```
site-packages\kubernetes\stream\ws_client.py in update(self, timeout)
    176            elif op_code == ABNF.OPCODE_BINARY or op_code == ABNF.OPCODE_TEXT:
    177                data = frame.data
    178                 if six.PY3:
--> 179                     data = data.decode("utf-8")
    180                 if len(data) > 1:

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd4 in position 1: unexpected end of data
```

berlinsaint:
the same error here. I solve it by change to `data = data.decode("utf-8", "replace")`

but. when the file is encode by other charset, this code is not used.  any other solutions?

roycaihw:
> when the file is encode by other charset, this code is not used. any other solutions?

from the [rfc](https://tools.ietf.org/html/rfc6455#section-5.6) we should decode uft8 for `OPCODE_TEXT` but leave `OPCODE_BINARY` as raw. this client should not assume encoding for arbitrary binary data

leavest:
> the same error here. I solve it by change to `data = data.decode("utf-8", "replace")`
> 
> but. when the file is encode by other charset, this code is not used. any other solutions?

I get the same error when use vim command in web terminal.

tbarrella:
https://github.com/kubernetes-client/python-base/commit/15474efbaf906bf557c4a38392a0b06c95ce7841 helped to fix the error above, but it seems like this issue should be re-opened with the intent of being able to stream bytes without trying to decode them

PR #104, opened 2018-11-26
decode add a replace option

saberuster:
as mentioned above #88 . 

we can determine whether data is utf8 encode in another way，instead of raise a exception on reading stream.

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -176,7 +176,7 @@ class WSClient:
             elif op_code == ABNF.OPCODE_BINARY or op_code == ABNF.OPCODE_TEXT:
                 data = frame.data
                 if six.PY3:
-                    data = data.decode("utf-8")
+                    data = data.decode("utf-8", "replace")
                 if len(data) > 1:
                     channel = ord(data[0])
                     data = data[1:]


saberuster:
I signed it

saberuster:
I signed it

saberuster:
I signed it

dbazhal:
can anyone approve this?

dbazhal:
Btw, don't know if it is correct to discuss here - i know that 'replace' is kind of solution, but it seems to me that it is incorrect to decode chunks of data without preliminary concatenation. For me it looks like chunk can really be broken - if it's bytes it's not necessary complete information about encoded symbol, but it may be just the first byte of two bytes total. In that case adding 'replace' could just corrupt data received by client.

saberuster:
i am agree with your. but in my scene：
if i exec `kubectl exec mypod cat /bin/my_binary` i can recieve some data like this:
```shell
\����̋���܍��,  
```
if i use python-base , i get an exception when reading data. i want to do same things with `kubectl ...` 
What do you think about this?

dbazhal:
> if i exec `kubectl exec mypod cat /bin/my_binary` i can recieve some data like this:
> ```shell
> \����̋���܍��,  
> ```
> What do you think about this?
> @dbazhal

I beleive kubectl throws raw binary data to output, but I'm not sure tries to decode it partially. I assume correct solution to this wold be to merge chunks, then try to decode it with 'replace'. But I'm not sure, that's just guess, need to dig deeper -)

roycaihw:
from the [rfc](https://tools.ietf.org/html/rfc6455#section-5.6) I think we should decode utf8 if opcode is `Text` and return raw data if opcode is `Binary`

the previous implementation seems to be modeled from WebSocketApp but I don't see why we assumed utf8 for `OPCODE_BINARY` https://github.com/kubernetes-client/python/pull/125#discussion_r101837622. cc @mbohlool

dbazhal:
> from the [rfc](https://tools.ietf.org/html/rfc6455#section-5.6) I think we should decode utf8 if opcode is `Text` and return raw data if opcode is `Binary`
> 
> the previous implementation seems to be modeled from WebSocketApp but I don't see why we assumed utf8 for `OPCODE_BINARY` [kubernetes-client/python#125 (comment)](https://github.com/kubernetes-client/python/pull/125#discussion_r101837622). cc @mbohlool

ok. as I understand, we should mark whole channel by the first frame opcode to treat all it's parts the same way? opcode isn't supposed to change fron frame to frame, right?

roycaihw:
I checked the websocket-client module that [we depend on](https://github.com/kubernetes-client/python-base/blob/1d5231cee6d0674b958ad429b60dda39363a4dc1/stream/ws_client.py#L76). It looks like the module has handled fragment concatenation for us already

our ws client invokes `recv_data_frame`
https://github.com/kubernetes-client/python-base/blob/1d5231cee6d0674b958ad429b60dda39363a4dc1/stream/ws_client.py#L177

the websocket-client [extracts the data type (text or binary) from the first frame, and accumulates following continuation frames into the payload](https://github.com/websocket-client/websocket-client/blob/3c25814664fef5b78716ed8841123ed1c0d17824/websocket/_core.py#L349-L354). The client only [returns the ws message when the last fragment (FIN) is received](https://github.com/websocket-client/websocket-client/blob/3c25814664fef5b78716ed8841123ed1c0d17824/websocket/_abnf.py#L436-L447). Note that we've disabled [`fire_cont_frame` and `skip_utf8_validation`](https://github.com/websocket-client/websocket-client/blob/3c25814664fef5b78716ed8841123ed1c0d17824/websocket/_core.py#L94-L95) already. 

I don't think we need one more layer to concatenate multiple ws messages. websocket-client guarantees each text-type ws message [can be decoded as utf8](https://github.com/websocket-client/websocket-client/blob/3c25814664fef5b78716ed8841123ed1c0d17824/websocket/_abnf.py#L443-L445). It's just we shouldn't assume binary-type ws messages can be decoded in the same way

dbazhal:
@roycaihw ok, that's great! Seems like there is no trouble with ws anymore.
But I think there is still trouble with watch. I tried to switch from my fork to kubernetes-9.0.0 and again got that same error
```
<...>
    seg = seg.decode('utf8')
```
So it looks like this concrete PR is outdated, but could you please take a look at https://github.com/kubernetes-client/python-base/pull/112 especially https://github.com/kubernetes-client/python-base/pull/112/files#diff-7768487132d1edef2f3cb4e8d4101890 ?

dbazhal:
@roycaihw I created separate PR on that https://github.com/kubernetes-client/python-base/pull/137

roycaihw:
This PR fixes a bug (not honoring opcode when decoding) in our ws client. I haven't convinced myself if we want to decode binary message using utf8, but this change is still better than failing

/lgtm
/approve

@dbazhal Watch doesn't use ws so that error should be unrelated. I will review https://github.com/kubernetes-client/python-base/pull/138

PR #105, opened 2018-12-02
Improve Empty auth-provider/config error message

iamneha:
Fixes issue https://github.com/kubernetes-client/python/issues/328

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -452,12 +452,12 @@ class ConfigNode(object):
 
     def __getitem__(self, key):
         v = self.safe_get(key)
+        if isinstance(v, dict) or isinstance(v, list):
+            return ConfigNode('%s/%s' % (self.name, key), v)
         if not v:
             raise ConfigException(
                 'Invalid kube-config file. Expected key %s in %s'
                 % (key, self.name))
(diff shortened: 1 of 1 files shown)


dims:
@iamneha it would be good to add a test case too

config/kube_config.py
```diff
@@ -452,12 +452,12 @@ def safe_get(self, key):
 
     def __getitem__(self, key):
         v = self.safe_get(key)
+        if isinstance(v, dict) or isinstance(v, list):
+            return ConfigNode('%s/%s' % (self.name, key), v)
```
roycaihw: Could you add some documentation to explain how this fixes https://github.com/kubernetes-client/python/issues/328? Adding a test will definitely help.

iamneha: Yeah sure I'll add test for this

micw523: I think the trick is to determine if the object is a dict type first - if it’s an empty dict object, it will be passed into this clause instead of being passed into the exception clause since the empty dict is false.

cblecker:
ping on this

cblecker:
@iamneha are you still interested in working on this PR?

iamneha:
@cblecker I am not working on this right now. Occupied with some other work

PR #107, opened 2018-12-03
Verify-boilerplate script

iamneha:
Fixes issue https://github.com/kubernetes-client/python/issues/22

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 # Copyright 2016 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/hack/boilerplate/boilerplate.py
+++ b/hack/boilerplate/boilerplate.py
@@ -1,0 +1,197 @@
+#!/usr/bin/env python
+
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+import argparse
+import datetime
+import difflib
+import glob
+import os
+import re
+import sys
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "filenames",
+    help="list of files to check, all files if unspecified",
+    nargs='*')
+
+rootdir = os.path.dirname(__file__) + "/../../"
+rootdir = os.path.abspath(rootdir)
+parser.add_argument(
+    "--rootdir", default=rootdir, help="root directory to examine")
+
+default_boilerplate_dir = os.path.join(rootdir, "hack/boilerplate")
+parser.add_argument(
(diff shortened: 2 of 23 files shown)


run_tox.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env python
```
dims: This is a bash script :)

dims:
/assign @mbohlool 

Mehdy, Roy, this looks ready

hack/verify-boilerplate.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Copyright 2014 The Kubernetes Authors.
```
micw523: It’s now 2018 ;)

hack/boilerplate/boilerplate.py
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env python
+
+# Copyright 2015 The Kubernetes Authors.
```
micw523: Same here

hack/boilerplate/boilerplate.py
```diff
@@ -0,0 +1,209 @@
...
+
+    for filename in filenames:
+        if not file_passes(filename, refs, regexs):
+            print(filename, file=sys.stdout)
+
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
micw523: Please add this file to the list of files for pycodestyle checking

iamneha: I'll add this file inside pycodestyle script inside kubernetes-client/python/script

config/__init__.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
```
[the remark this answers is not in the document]
dims: @micw523 this seems to be what was in k/k main repo for boilerplate. i think this is ok - https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.py.txt#L1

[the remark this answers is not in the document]
micw523: @dims I agree if that’s how the main repo did things. Since I’m not very familiar with the main repo, does the main repo contain any python functions or it’s all executables with main()?

hack/boilerplate/boilerplate.py
micw523: I’m fine with it but please run autopep8 with aggressive level 2 on your file to check for the coding style to see if there’s any changes.

config/__init__.py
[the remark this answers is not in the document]
roycaihw: Does the script only verify python-base files? I'd expect CI to fail since we don't have the python shebang in most of the generated files in [python](https://github.com/kubernetes-client/python) repo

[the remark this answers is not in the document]
micw523: @roycaihw correct me if I’m wrong, but the python files in the main repo seem to be mostly executables?

hack/verify-boilerplate.sh
dims: +1 to fix up the date

hack/boilerplate/boilerplate.py
dims: @micw523 i think we should leave it as a simple copy of the code in main k/k repository. If we need to do more, we should do there first and copy it over here.

mbohlool:
This is generally looks good to me. However, I didn't get the point about python-base vs the main python repo. If this is only run for the python-base, I couldn't see why in 2 minutes I spend on this. I defer the review decision to @roycaihw as he had the same comment.

dims:
@mbohlool let's start here and we can even reuse in the other repo. for me the motivation was that there were at least 3 files which had some bits in the license header wrong

roycaihw:
/approve

I think verifying python-base on its own is helpful

Issue #54, opened 2018-03-28
userinfo.email scope is needed to work with rbac


jlewi:
[KubeConfigLoader._refresh_credentials](https://github.com/kubernetes-client/python-base/blob/master/config/kube_config.py#L137) sets the scope to cloud-platform.

I think userinfo.email scope is needed as well.

See kubernetes/kubernetes#58141 for a similar issue with kubectl and a good explanation.

I think we need the userinfo.email scope because RBAC rules can be expressed in terms of the email of service accounts. But if the userinfo.email scope isn't included that APIServer ends up using the numeric id of service accounts which won't work if RBAC rules are written in terms of 
the emails.

I haven't confirmed for myself this is an issue (I'm working through a variety of issues with kubectl/kubeconfig/client libs) so I could be wrong.

axelsteingrimsson:
I believe we are experiencing the exact issue that you are describing when using the [Kubernetes Python client](https://github.com/kubernetes-client/python). We are finding that when the Python client refreshes the access token in `~/.kube/config` we can't make requests to the Kubernetes API server with a service account that is normally authenticated with RBAC. Though when we use `kubectl` the token refresh is fine.

Using a `~/.kube/config` with the `users` section of the format:
```
users:
- name: application-default-credentials
  user:
    auth-provider:
      config:
        access-token: <REDACTED_ACCESS_TOKEN>
        expiry: '2018-11-11T11:11:11Z'
      name: gcp
```

And a GCP service account (not a user service account, but a server service account) with the format:
```
{
  "type": "service_account",
  "project_id": "example-project-id",
  "private_key_id": "<REDACTED_PRIVATE_KEY_ID>",
  "private_key": "<REDACTED_PRIVATE_KEY>",
  "client_email": "<REDACTED_CLIENT_EMAIL>",
  "client_id": "REDACTED_CLIENT_ID",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://accounts.google.com/o/oauth2/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/<REDACTED_CLIENT_EMAIL>"
}
```

I found that if I used `kubectl` with this configuration and service account to refresh the access token then it would be valid and I'm able to make requests to the Kubernetes API server. However, if
[...]

PR #110, opened 2018-12-12
Add email scope to GCP provided credential refresh

axelsteingrimsson:
This would fix the issue described in: https://github.com/kubernetes-client/python-base/issues/54

This is similar to the PR to [fix the issue with refreshing the service account token](https://github.com/kubernetes-client/python-base/pull/40), except this is specifically targeted at addressing service account tokens that are authenticated using RBAC. 

The issue is essentially that if the service account that is making a request to the Kubernetes API server and the request is authenticated using RBAC, then the access token will be invalid because it isn't signed by an email. By passing the `"https://www.googleapis.com/auth/userinfo.email"` when creating the credentials it will generate an access token that can be authenticated using RBAC.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -141,9 +141,10 @@ class KubeConfigLoader(object):
         self._config_persister = config_persister
 
         def _refresh_credentials():
-            credentials, project_id = google.auth.default(
-                scopes=['https://www.googleapis.com/auth/cloud-platform']
-            )
+            credentials, project_id = google.auth.default(scopes=[
+                'https://www.googleapis.com/auth/cloud-platform',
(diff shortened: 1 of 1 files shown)


axelsteingrimsson:
Hey there @mbohlool and @yliaog, (sorry for the ping) it's no big rush, but would you guys mind taking a look at this if you have any time soon? :)

yliaog:
thanks for the pr. so you have tested it manually that it works, right? just want to double confirm.

axelsteingrimsson:
Yes, we've been running a patched version of this client with this change for the past few months in production and we haven't seen any adverse behaviour.

PR #112, opened 2019-01-21
fix stream data decoding

dbazhal:
Related to https://github.com/kubernetes-client/python-base/issues/88 and https://github.com/kubernetes-client/python-base/pull/104
I suppose decoding complete data is better than decoding data chunks.
Would be nice to get review from @mbohlool  and @saberuster

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -53,7 +53,8 @@ class WSClient:
             header.append("authorization: %s" % headers['authorization'])
 
         if headers and 'sec-websocket-protocol' in headers:
-            header.append("sec-websocket-protocol: %s" % headers['sec-websocket-protocol'])
+            header.append("sec-websocket-protocol: %s" %
+                          headers['sec-websocket-protocol'])
         else:
             header.append("sec-websocket-protocol: v4.channel.k8s.io")
 
@@ -186,8 +187,8 @@ class WSClient:
                     data = data[1:]
                     if data:
                         if channel in [STDOUT_CHANNEL, STDERR_CHANNEL]:
-                            # keeping all messages in the order they received for
-                            # non-blocking call.
+                            # keeping all messages in the order they received
+                            # for non-blocking call.
                             self._all += data
                         if channel not in self._channels:
                             self._channels[channel] = data
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -1,14 +1,18 @@
-# Licensed under the Apache License, Version 2.0 (the "License"); you may
-# not use this file except in compliance with the License. You may obtain
-# a copy of the License at
+#!/usr/bin/env python
+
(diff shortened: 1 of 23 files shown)


dbazhal:
Ok, i think I now signed the cla, hope the bot will reconcile that.

dbazhal:
@roycaihw or @micw523 or @yliaog guys, mb any of you could take a look at my pr, or https://github.com/kubernetes-client/python-base/pull/104 please? don't want to patch this in all my controllers, and with current amount of objects in crds they fail with unicode decode errors like every hour :D

stream/ws_client.py
```diff
@@ -98,15 +107,21 @@ def readline_channel(self, channel, timeout=None):
         while self.is_open() and time.time() - start < timeout:
             if channel in self._channels:
                 data = self._channels[channel]
-                if "\n" in data:
-                    index = data.find("\n")
+                newline_symbol = "\n"
+                if six.PY3:
+                    newline_Symbol = b"\n"
```
micw523: Although it does the same thing, I wonder why you applied different logic compared to [Line 44](https://github.com/kubernetes-client/python-base/pull/112/files#diff-eb24b95d54baa73740535635b5f56b5aR44). Also, be careful about your variable name: do you mean ```newline_symbol``` instead of a capitalized S?

micw523:
Also, it doesn't seem like your CLA has been updated yet. You sure you signed it?

stream/ws_client.py
dbazhal: uh-oh, my bad, fixed that. i'll this code tomorrorow. 
and it'd be great if you could judge the solution to decoding errors itself, it bothers me more than anything else :) maybe you will find solution from https://github.com/kubernetes-client/python-base/pull/104 better than this one

roycaihw:
commented in https://github.com/kubernetes-client/python-base/pull/104#issuecomment-460426918. I think we should decode utf8 only if opcode is `Text`

dbazhal:
@roycaihw I've added decoding on opcode_text. Not sure if this last change covers all the cases, but maybe you already could review this solution?

dbazhal:
Hey guys, what else can I do to help with this one? @roycaihw ?

dbazhal:
I dunno if it is acceptable to up threads, but it would be nice to get any answer ^_^

roycaihw:
@dbazhal sorry for the late reply. I will review this pull this week

dbazhal:
just another firendly reminder :)

roycaihw:
@dbazhal my apologies.. was busy with k/k 1.15 freeze. Please take a look at https://github.com/kubernetes-client/python-base/pull/104#issuecomment-501095047. I think `recv_data_frame` handles fragment concatenation already, so we don't have to concatenate the messages on top of that

PR #122, opened 2019-03-14
Don't fail if tmp files have been deleted

zach-nervana:
I had a case where files in `/tmp` were deleted I get the exception on line 118: `File does not exists: /tmp/xxxxx`. If we are going to catch it, we may as well recover from it. Especially since the file is just a cache.


tomplus:
It's known issue, please take a look at #38

PR #124, opened 2019-03-21
Strip trailing slashes from `idp-issuer-url` to avoid 404 on refreshing token

andyhd:
My `idp-issuer-url` ends with a trailing slash for some reason, and `KubeConfigLoader._refresh_oidc` concatenates it with `/.well-known/openid-configuration` to generate a URL that 404s.

I've just stripped the trailing slash to fix this.


andyhd:
I've signed the CLA

roycaihw:
please add a test case for this fix

PR #127, opened 2019-04-05
Drop Python3.4

oz123:
Python 3.4 has become EOL on March 18, 2019.

 https://www.python.org/dev/peps/pep-0429/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ matrix:
        env: TOXENV=docs
      - python: 2.7
        env: TOXENV=coverage,codecov
-     - python: 3.4
-       env: TOXENV=py34
      - python: 3.5
        env: TOXENV=py35
      - python: 3.5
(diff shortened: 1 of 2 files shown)


yliaog:
3.4.10 final: March 18, 2019

the last release was 3/18/2019.
I think it's better to wait for a bit longer before dropping it.

micw523:
/hold
Let's keep this open and merge this after a few months then.

oz123:
@yliaog can this be moved forward. It's not a good service for the community to support python3.4.

yliaog:
@micw523 what do you think?

micw523:
@yliaog It's probably fine. I see Matplotlib has already dropped Python 3.4.
/cc @roycaihw

yliaog:
@micw523 sounds good to me. could you please cancel the hold then?

PR #129, opened 2019-04-20
Fix python3 UnicodeDecodeError when use websocket

### LINKED ISSUE #88: stream assumes all data as UTF-8. That's not always the case.

leavest:
Fixes kubernetes-client/python/issues#717 
Fixes #88 
in python3

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -24,6 +24,7 @@ from websocket import WebSocket, ABNF, enableTrace
 import six
 import ssl
 from six.moves.urllib.parse import urlencode, quote_plus, urlparse, urlunparse
+import chardet
 
 STDIN_CHANNEL = 0
 STDOUT_CHANNEL = 1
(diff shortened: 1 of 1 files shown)


leavest:
reply

leavest:
I signed it

PR #130, opened 2019-04-20
Fix stream decode error 

### LINKED ISSUE #88: stream assumes all data as UTF-8. That's not always the case.

leavest:
Fix #88 .
#104 and #112 fix the same bug.
I work well with is.

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -24,6 +24,7 @@ from websocket import WebSocket, ABNF, enableTrace
 import six
 import ssl
 from six.moves.urllib.parse import urlencode, quote_plus, urlparse, urlunparse
+import chardet
 
 STDIN_CHANNEL = 0
 STDOUT_CHANNEL = 1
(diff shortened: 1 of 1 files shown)


leavest:
The same bug is in https://github.com/kubernetes-client/python/issues/717.
Not sure if this  covers all the cases but it worked for some people.

leavest:
I find bug in this code, close it

PR #132, opened 2019-04-30
Ensure no base64 incorrect padding errors occur in PY3

ralphje:
Using a Kubernetes oid token configuration with Python 3, I encountered incorrect padding errors when the library attempted to decode the base64 string. Looking into the code base, I noticed that == is added to avoid this error in Python 2, but is mysteriously omitted from the Python 3 call.

I'm not sure why this is, especially since Python 3 has the exact same behaviour regarding insufficient padding as Python 2. Since the trick with adding == still works in Python 3, I have simply copied this from the Python 2 implementation and added a comment why this magic constant works.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -282,9 +282,11 @@ class KubeConfigLoader(object):
         if len(parts) != 3:  # Not a valid JWT
             return None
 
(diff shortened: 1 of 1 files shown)


ralphje:
Appears to duplicate #79, though my fix was a more simple approach.

PR #133, opened 2019-05-12
Retry expired watches

### LINKED ISSUE #57: Watch stream should handle HTTP error before unmarshaling event

mitar:
This is a followup to #102 and fixes #57.

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import http
 import json
 import pydoc
 
@@ -86,7 +87,7 @@ class Watch(object):
     def unmarshal_event(self, data, return_type):
         js = json.loads(data)
         js['raw_object'] = js['object']
-        if return_type:
+        if return_type and js['type'] != 'ERROR':
             obj = SimpleNamespace(data=json.dumps(js['raw_object']))
             js['object'] = self._api_client.deserialize(obj, return_type)
             if hasattr(js['object'], 'metadata'):
@@ -101,6 +102,14 @@ class Watch(object):
 
     def stream(self, func, *args, **kwargs):
         """Watch an API resource and stream the result back via a generator.
+
+        Note that watching an API resource can expire. The method tries to
+        resume automatically once from the last result, but if that last result
+        is too old as well, an `ApiException` exception will be thrown with
+        ``code`` 410. In that case you have to recover yourself, probably
+        by listing the API resource to obtain the latest state and then
+        watching from that state on by setting ``resource_version`` to
+        one returned from listing.
 
         :param func: The API function pointer. Any parameter to the function
                      can be passed after this parameter.
@@ -134,6 +143,7 @@ class Watch(object):
             self.resource_version = kwargs['resource_version']
 
         timeouts = ('timeout_seconds' in kwargs)
+        retry_after_410 = False
         while True:
             resp = func(*args, **kwargs)
             try:
@@ -141,7 +151,23 @@ class Watch(object):
                     # unmarshal when we are receiving events from watch,
                     # return raw string when we are streaming log
                     if watch_arg == "watch":
-                        yield self.unmarshal_event(line, return_type)
+                        event = self.unmarshal_event(line, return_type)
+                        if isinstance(event, dict) \
+                                and event['type'] == 'ERROR':
+                            obj = event['raw_object']
+                            # Current request expired, let's retry,
+                            # but only if we have not already retried.
+                            if not retry_after_410 and \
+                                    obj['code'] == http.HTTPStatus.GONE:
+                                retry_after_410 = True
+                                break
+                            else:
+                                reason = "%s: %s" % (obj['reason'], obj['message'])
+                                raise client.rest.ApiException(status=obj['code'],
+                                                               reason=reason)
+                        else:
+                            retry_after_410 = False
+                            yield event
                     else:
                         yield line
                     if self._stop:
--- a/watch/watch_test.py
+++ b/watch/watch_test.py
@@ -16,6 +16,8 @@ import unittest
 
 from mock import Mock, call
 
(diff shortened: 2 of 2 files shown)


ccamacho:
It would be nice to have those 3 commits squashed into 1

mitar:
Doesn't GitHub support that when merging the PR?

micw523:
Manually yes, but we’re using bots to merge the commits so usually the commits need to be squashed.

mitar:
Done.

watch/watch.py
```diff
@@ -138,11 +146,26 @@ def stream(self, func, *args, **kwargs):
...
         timeouts = ('timeout_seconds' in kwargs)
+        retry_after_410 = False
         while True:
             resp = func(*args, **kwargs)
             try:
                 for line in iter_resp_lines(resp):
-                    yield self.unmarshal_event(line, return_type)
+                    event = self.unmarshal_event(line, return_type)
+                    if isinstance(event, dict) and event['type'] == 'ERROR':
+                        obj = event['raw_object']
```
micw523: Is it safe to assume the APIs won't change for accessing ```event['raw_object']['code']```, @roycaihw ?

watch/watch.py
```diff
@@ -138,11 +146,26 @@ def stream(self, func, *args, **kwargs):
...
+                        # Current request expired, let's retry,
+                        # but only if we have not already retried.
+                        if not retry_after_410 and obj['code'] == 410:
+                            retry_after_410 = True
+                            break
```
roycaihw: we should not retry the watch request with the same RV when it's already gone in apiserver

we can retry on 500 and 504 according to the [client-go implementation](https://github.com/kubernetes/kubernetes/blob/270d9529e004f04829a7cc9e5c83e43160a44f3a/staging/src/k8s.io/client-go/tools/watch/retrywatcher.go#L203-L211)

watch/watch.py
```diff
@@ -91,7 +91,7 @@ def unmarshal_event(self, data, return_type):
         except ValueError:
             return data
         js['raw_object'] = js['object']
-        if return_type:
+        if return_type and js['type'] != 'ERROR':
```
roycaihw: this is where we fail to deserialize the response. Good catch

watch/watch.py
roycaihw: @yliaog pointed out that we need to be careful about the code structure here. We want to be consistent to kubernetes client-go in terms of what a watch is capable of doing

- In client-go, a watch function call [doesn't retry on err](https://github.com/kubernetes/kubernetes/blob/5219ad7be209b31cdfdae2eda1cd435aff4a3cd2/staging/src/k8s.io/apimachinery/pkg/watch/streamwatcher.go#L99). A client can implement their own handling for watch errors
- client-go provides tooling like informer and retrywatcher for clients to leverage, which have retry logics for watch errors

We should be consistent and provide the same choices to our user

dmateusp:
is this good to go ?

jinchihe:
Can someone review and approve the PR? we meet the problem now... Thanks.

watch/watch.py
mitar: But that would be backwards incompatible with current offering by Python package? It already supports retrying. So I think for Python it seems we already made decision that watch is retrying.

mel1nn:
Could you fix the merge conflict please ?

mel1nn:
@mitar ?

mitar:
Sorry. I will look into this now.

mitar:
Done.

mitar:
I have updated this. Please review it before it gets rotten again.

mel1nn:
@mbohlool ?

mel1nn:
@roycaihw ?

watch/watch.py
```diff
@@ -134,14 +142,30 @@ def stream(self, func, *args, **kwargs):
...
                     # return raw string when we are streaming log
                     if watch_arg == "watch":
-                        yield self.unmarshal_event(line, return_type)
+                        event = self.unmarshal_event(line, return_type)
+                        if isinstance(event, dict) \
+                                and event['type'] == 'ERROR':
+                            obj = event['raw_object']
+                            # Current request expired, let's retry,
+                            # but only if we have not already retried.
+                            if not retry_after_410 and obj['code'] == 410:
```
yliaog, line 159: better to use the code in http.HTTPStatus to avoid using magic const 410

mitar: Updated.

watch/watch.py
```diff
@@ -102,6 +103,14 @@ def unmarshal_event(self, data, return_type):
     def stream(self, func, *args, **kwargs):
         """Watch an API resource and stream the result back via a generator.
 
+        Note that watching an API resource can expire. The method tries to
+        resume automatically from the last result, but if that last result
```
yliaog, line 107: i think this comment has already implied it, but still better to make it clear that only *retry once*, so user has the correct expectation on how many times retry is attempted.

yliaog:
thanks for the PR.

/lgtm
/approve

PR #134, opened 2019-05-13
Update _load_azure_token

rabidang3ls:
_load_azure_token would fail if the "expires-on" field was not comprised of digits. The field isn't guaranteed to be digits, so this updates the code to also accept the date format `%Y-%m-%d %H:%M:%S.%f

FIx for #84

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -227,8 +227,12 @@ class KubeConfigLoader(object):
         if 'access-token' not in provider['config']:
             return
         if 'expires-on' in provider['config']:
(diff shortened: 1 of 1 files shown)


bitsofinfo:
@rabidang3ls whats the status of this? really need it! thank you

rabidang3ls:
@bitsofinfo I don't know the project well enough to understand their travis-ci output. Not sure what fails, but I haven't had any issues with it. Looks like #141 passed all checks and should be implementing the same fix.

PR #137, opened 2019-06-20
Fix data decode for watch/watch.py:49

dbazhal:
There is error like 
```
<...>
    seg = seg.decode('utf8')
```
So this PR fixes that. Related somehow to https://github.com/kubernetes-client/python-base/pull/104

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -16,6 +16,7 @@
 
 import json
 import pydoc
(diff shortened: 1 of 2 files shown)


dbazhal:
Much sorry, messed up with commit author and git history... Will recreate this PR again.

PR #138, opened 2019-06-20
Fix data decode for watch/watch.py:49

dbazhal:
There is error like 
```
<...>
    seg = seg.decode('utf8')
```
So this PR fixes that. Related somehow to https://github.com/kubernetes-client/python-base/pull/104

### DIFF
--- a/watch/watch_test.py
+++ b/watch/watch_test.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import six
 import unittest
 
 from mock import Mock, call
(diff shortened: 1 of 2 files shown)


watch/watch_test.py
```diff
@@ -40,6 +41,17 @@ def test_watch_with_decode(self):
...
+            fake_resp.read_chunked = Mock(
+                return_value=[
+                    b'{"type": "ADDED", "object": {"metadata": {"name": "test1",'
+                    b'"resourceVersion": "1"}, "spec": {}, "status": {}}}\n',
+                    b'{"type": "ADDED", "object": {"metadata": {"name": "test2",'
+                    b'"resourceVersion": "2"}, "spec": {}, "sta',
+                    b'tus": {}}}\n'
+                    b'{"type": "ADDED", "object": {"metadata": {"name": "test3",'
+                    b'"resourceVersion": "3"}, "spec": {}, "status": {}}}\n',
+                    b'should_not_happened\n'])
```
roycaihw: I'm confused. Does the Python 3 client always get byte formatted data in a watch response? If so, does that mean the Python 3 client never supported watch? 

I'm not sure if I understand what triggers a byte formatted response from the apiserver. Could you add an e2e test in https://github.com/kubernetes-client/python/blob/master/kubernetes/e2e_test/test_client.py for watch, so that we can see watch works for both Python 2 and 3?

dbazhal: I'm not sure, but it seems to me that yes, p3 always gets bytes. I'm not sure about 'never supported watch', I may guess that it's not working only if data contains non-ascii symbols.

Writing e2e test is a great idea, but I'm not quiet sure I understand how watch test should look like. If you could help me out with describing test case, I'd do it faster :)

In testcase test_watch I'm creating, let's say, configmap, remember it's revision, make changes to this configmap, and start watching it from saved revision, sounds legit?

PR #140, opened 2019-06-26
Check if its possible to parse a chunked response

### LINKED ISSUE #139: Python kubernetes client breaks when parsing the response from custom controller
Issue: Python kubernetes client breaks when parsing the response from custom controller

When running a watch primitive as:
    resource_version = ''
    while True:
        stream = watch.Watch().stream(crds.list_cluster_custom_object, DOMAIN, "v1", "guitars", resource_version=resource_version)

The python client breaks with:

  File "controller-pods.py", line 39, in <module>
    for event in stream:
  File "/usr/lib/python2.7/site-packages/kubernetes/watch/watch.py", line 154, in stream
  File "/usr/lib/python2.7/site-packages/kubernetes/watch/watch.py", line 60, in iter_resp_lines
    resp = resp.read_chunked(decode_content=False)
AttributeError: 'HTTPResponse' object has no attribute 'read_chunked'

Packages versions:
rpm -qf /usr/lib/python2.7/site-packages/openshift/dynamic/client.py /usr/lib/python2.7/site-packages/kubernetes/watch/watch.py

    python2-openshift-0.8.8-1.el7.noarch
    python2-kubernetes-8.0.1-1.el7.noarch

Workaround, it seems that in this case the response does not allow to be chunked, using instead **read() in[1]:

https://github.com/kubernetes-client/python-base/blob/master/watch/watch.py#L48

Before:
    for seg in resp.read_chunked(decode_content=False):
After:
    for seg in resp.read():

Seemed to work fine.

ccamacho:
This patch is a simple check for either read the chunks
of the response of use the standard read method.
Currently if it is not possible to use the 'read_chunked'
method the client fails.
This commit fixes #139

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -45,10 +45,27 @@ def _find_return_type(func):
 
 def iter_resp_lines(resp):
     prev = ""
-    for seg in resp.read_chunked(decode_content=False):
+    # Check if it is possible to read a chunked
+    # response, if not, use the standard read.
+    try:
+        resp = resp.read_chunked(decode_content=False)
+        for seg in resp:
+            if isinstance(seg, bytes):
+                seg = seg.decode('utf8')
+            seg = prev + seg
+            lines = seg.split("\n")
+            if not seg.endswith("\n"):
+                prev = lines[-1]
+                lines = lines[:-1]
+            else:
+                prev = ""
+            for line in lines:
+                if line:
+                    yield line
+    except AttributeError:
+        seg = resp.read()
         if isinstance(seg, bytes):
             seg = seg.decode('utf8')
-        seg = prev + seg
         lines = seg.split("\n")
         if not seg.endswith("\n"):
             prev = lines[-1]


ccamacho:
Hey folks @mbohlool @roycaihw, I'm hitting all over #139 and I wanted to know if this can be done. I already have patched my clusters and seemed to work fine.

Thanks!!

ccamacho:
@mmazur @tripledes reviews are welcomed :)

mmazur:
Right, so let me start with the fact that I'm not familiar with any of this code and I'm not in a position to merge anything.

However, from doing some digging around it looks to me that the proper way to handle chunky responses is not with exceptions, but by emulating what the authoritative source (urllib3 itself) does. Which is [this](https://github.com/urllib3/urllib3/blob/master/src/urllib3/response.py#L557). Now, I'm not sure what the `is_fp_closed()` thing does and whether it's applicable in our case, but it looks to me like `if self.chunked and self.supports_chunked_reads()` is what we want to be doing here in lieu of catching exceptions.

ccamacho:
@mmazur yeahp, you are actually right as we can handle the chunked case in that way. The main problem is that the urllib3 version we deliver does not have those methods.

rpm -qf /usr/lib/python2.7/site-packages/urllib3/response.py
python-urllib3-1.10.2-5.el7.noarch

Presently from 1.25.3

That's the main reason I decided to use an exception instead, just to avoid the dependency.

roycaihw:
may I ask what responses from Kubernetes APIs do not support read_chunked?

ccamacho:
Hi @roycaihw thanks a lot for your reply! And sorry for not replying faster.
The issue in question is described in #139, basically when calling the watch() method the client breaks.

rpm -qf /usr/lib/python2.7/site-packages/urllib3/response.py
python-urllib3-1.10.2-5.el7.noarch

For example:
stream = watch.Watch().stream(crds.list_cluster_custom_object, DOMAIN, "v1", "guitars", resource_version=resource_version)


The problem is that the urllib version we ship does not have this method read_chunked, and handling it with an exception allows having backwards compatibility with older versions of urllib.

The fix is actually really simple and we are already using this fix but it would be awesome if we can include it by default as it affects any OpenShift and OKD distribution based in the 4.0 version.

Thanks!

PR #141, opened 2019-07-03
Fix issue 84: Update _load_azure_token to hande str and int

rogerhmar:
Fix bug by first checking if it is a digit

Add tests for:
* time str
* int
* illegal str
* illegal int

There were several pull requests on this, but none seem to have tests.

This should fix https://github.com/kubernetes-client/python-base/issues/84

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -226,7 +226,8 @@ class KubeConfigLoader(object):
         if expires_on.isdigit():
             return int(expires_on) < time.time()
         else:
-            return time.strptime(expires_on, '%Y-%m-%d %H:%M:%S.%f') < time.gmtime()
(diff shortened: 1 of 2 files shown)


rogerhmar:
@roycaihw and @yliaog: Do you have the time to look at this?

Sephtex:
For which version of the kubernetes pip package will this fix be available? I need the openshift-0.9.0 package which I have installed and it has kubernetes-9.0.0 as dependency. I updated kubernetes to v10.0.0 and I couldn't find the fix yet. I suppose this fix will land in 9.0.0?

yliaog:
it will land in the next release. the current release is  v10.0.0. so v11.0.0 should have it

Issue #106, opened 2018-12-03
select should not be used in WSClient because python-websockets does its own buffering


PaulFurtado:
I'm hitting an issue where some exec calls made via the python client hang.

For background, I'm in the process of migrating my cluster from docker to cri-o and I have a suite of acceptance tests written in python that I run against the cluster to verify functionality. A good deal of of the tests exec commands inside pods and upon switching to cri-o, a certain command in the tests hangs/times out every time. Initially, I thought there was a bug in cri-o, but If the same command is executed via kubectl, it does not hang, so I started diving into the python client.

I believe I've tracked the problem down to the select call that is made in WSClient: https://github.com/kubernetes-client/python-base/blob/67a69d46560c4cf12e25266c41417f0dc0cc81ce/stream/ws_client.py#L169-L170
If I remove that select call, commands never hang.

I think that it's invalid to call select there because select is a system call that checks if a socket ready to be read/written, however, with websockets, we're several layers of abstraction away from the underlying system socket that select is checking and buffering is occurring at each of the layers. I suspect what is happening is:
1. recv_data_frame is called. The underlying recv call on the socket uses a fixed buffer size and receives multiple frames and recv_data_frame returns only one of them.
2. on the next iteration, we check if the underlying socket has any bytes available using select. It does not, but there is a frame waiting in the buffer in websocket-client that recv_data_frame would instantly return.

So I think the solution is to simply remove that select call, but I'm curious why it was added in the first place and whether removing it breaks some expectation. The only thing that I can think of is that it's being used to make read timeouts possible, however, if so, that is still buggy because the select could return if 1 byte is available to read, but then the recv calls in websocket-client would still block waiting on a complete frame.

@mbohlool what do you think?

bboe:
To update on this issue, I ran into an issue with select as documented here: https://github.com/kubernetes-client/python-base/issues/53#issuecomment-494094032

My thought was to convert to poll (has potential problems as not all operating systems support poll (but maybe all that support this package do?). However, if this select isn't needed, then that simplifies things even more.

PaulFurtado:
@bboe are you talking about epoll? If so, epoll has all the same problems as select here: the underlying socket's readiness is not equivalent to the websocket's readiness as the websocket is a logical construct with its own framing which the operating system doesn't know about

bboe:
[...]

PR #143, opened 2019-07-04
stream, ws: disallow timeout<0 when provided timeout is >=0

TomasTomecek:
Fixes https://github.com/kubernetes-client/python/issues/602
Fixes https://github.com/kubernetes-client/python-base/issues/106

All the credits go to @PaulFurtado who discovered the solution - I'm just trying to upstream it.

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -100,7 +100,10 @@ class WSClient:
         if timeout is None:
             timeout = float("inf")
         start = time.time()
-        while self.is_open() and time.time() - start < timeout:
+        while self.is_open():
+            remain = timeout - (time.time() - start)
+            if remain < 0:
+                return
             if channel in self._channels:
                 data = self._channels[channel]
                 if "\n" in data:
@@ -200,11 +203,14 @@ class WSClient:
         received during this time."""
         if timeout:
             start = time.time()
-            while self.is_open() and time.time() - start < timeout:
-                self.update(timeout=(timeout - time.time() + start))
+            while self.is_open():
+                remain = timeout - (time.time() - start)
+                if remain < 0:
+                    return
+                self.update(timeout=remain)
         else:
             while self.is_open():
-                self.update(timeout=None)
+                self.update()
 
     def close(self, **kwargs):
         """


PaulFurtado:
lgtm :+1:

TomasTomecek:
I logged in to the CLA portal since we, as Red Hat, signed the CLA but sadly it seems the bot is not picking it up

I also don't understand the CI failure(s): there are no errors.

micw523:
Looks like your change caused EXEC to hang and that's the failure.

roycaihw:
/cc

trying to catch up. The select was added in https://github.com/kubernetes-client/python/pull/125/files#diff-4ac213a679f24e620ff4d3e52b274287R160, which I believe models https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py. I'd like to understand why the ws-client package had it

PaulFurtado:
The thing that this PR breaks (further) is timeouts. In that, there is now no case where we will ever timeout waiting on a read.

I believe that that test has always relied on timeouts to succeed: https://github.com/kubernetes-client/python/blob/master/kubernetes/e2e_test/test_client.py#L94-L113
We should do some print debugging. It's possible it only hangs at line 112, where that update call should never read anything new.

That said, the websocket-client library offers no way to timeout a call to recv_data_frame: https://github.com/websocket-client/websocket-client/blob/master/websocket/_core.py#L333
Which explains why the timeout was hacked in using that invalid call to select. It appears to be not entirely trivial to modify websocket-client to add timeouts to individual reads, but it would be possible...

I'm having a hard time finding alternative websocket clients that work with python 2 and don't depend on running a twisted event loop :cry:. Anyone know of any?

TomasTomecek:
on the other hand, you can set timeout on the connection: https://github.com/websocket-client/websocket-client/blob/master/websocket/_socket.py#L49

the websocket-client is also calling select(): https://github.com/websocket-client/websocket-client/blob/master/websocket/_socket.py#L94

TomasTomecek:
I updated the PR to raise a TimeoutError exception when the remaining time is <= 0. This change should improve the situation, for more info see the related issues in my original post.

TomasTomecek:
I finally figured this out (thanks to my colleagues and a good lunch).

The problem is, that `timeout=(timeout - time.time() + start)` can result into timeout being `< 0` which means that select would be forever: the question is, how it's possible that such situation would happen - feels like an apiserver problem.

The other thing is that the websocket library actually does [select()](https://github.com/websocket-client/websocket-client/blob/master/websocket/_socket.py#L94) on its own, which on the other hand, shouldn't be a problem.

PaulFurtado:
Select is also a problem for others because you must pass it a file descriptor number under 1024, so in a big program it's trivial to exceed that. It only ever watches one file descriptor from select, so that means it's trivial to port from `select` to `poll`.

stream/ws_client.py
```diff
@@ -112,31 +115,31 @@ def readline_channel(self, channel, timeout=None):
                     else:
                         del self._channels[channel]
                     return ret
-            self.update(timeout=(timeout - time.time() + start))
```
roycaihw: > The problem is, that timeout=(timeout - time.time() + start) can result into timeout being < 0 which means that select would be forever: the question is, how it's possible that such situation would happen - feels like an apiserver problem.

I don't think it's in server side. We have `time.time() - start < timeout` before those timeout assignment. Doesn't this mean we have a race condition in between that allows negative `timeout`, which causes `select` to wait forever? 

If it is a race condition, I prefer we return on timeout instead of raise an error, since it's more backwards compatible

stream/ws_client.py
```diff
@@ -164,47 +167,47 @@ def write_stdin(self, data):
...
-    def update(self, timeout=0):
+    def update(self):
         """Update channel buffers with at most one complete frame of input."""
         if not self.is_open():
             return
         if not self.sock.connected:
             self._connected = False
             return
-        r, _, _ = select.select(
-            (self.sock.sock, ), (), (), timeout)
```
roycaihw: > Select is also a problem for others because you must pass it a file descriptor number under 1024, so in a big program it's trivial to exceed that. It only ever watches one file descriptor from select, so that means it's trivial to port from select to poll.

removing the `select` and asking existing users to wrap their own `select` or polling (e.g. https://github.com/kubernetes-client/python/pull/881) seems to be a big breaking change to me. 

cc @yliaog ^^^

Is the select hanging all the time or is this caused by a race condition in our timeout calculation?

TomasTomecek: I agree with you that the change in behaviour is not nice. Since the websocket lib does select() as well, I was hoping to get at least some consistency.

I can reproduce the hang consistently, it usually happens when megabytes of data flow through the exec interface (some buffer fills in?)

I'm gonna revert this change and try the one you propose above.

stream/ws_client.py
TomasTomecek: I'm trying to debug this even further and the problematic bit is that `websocket_call(...)` gets `_request_timeout = None` by default which results into infinite loop of:
```python
            while self.is_open():
                self.update()
```
I have no idea why the socket doesn't get closed.

TomasTomecek: But when I call stream like this:
```python
stream(
    self.api.connect_get_namespaced_pod_exec,
    self.pod_name,
    self.k8s_namespace_name,
    command=command,
    stdin=False,
    stderr=True,
    stdout=True,
    tty=False,
    _request_timeout=5.0
)
```

it suddenly works just fine and the code is no longer stuck in an infinite loop

stream/ws_client.py
```diff
@@ -100,7 +100,10 @@ def readline_channel(self, channel, timeout=None):
         if timeout is None:
             timeout = float("inf")
         start = time.time()
-        while self.is_open() and time.time() - start < timeout:
+        while self.is_open():
+            remain = timeout - (time.time() - start)
```
roycaihw: use `remain` in L118 as well? 
```
self.update(timeout=remain)
```

stream/ws_client.py
```diff
@@ -100,7 +100,10 @@ def readline_channel(self, channel, timeout=None):
...
+            if remain < 0:
```
roycaihw: also when `remain` is equal to 0?

stream/ws_client.py
roycaihw: > I'm trying to debug this even further and the problematic bit is that websocket_call(...) gets _request_timeout = None by default which results into infinite loop of:

Good catch! I tried locally and also found `_request_timeout = None`. This is problematic because I would expect it to be defaulted to 60
https://github.com/kubernetes-client/python-base/blob/95858d5ce843e98d65a327f3099cd0b6cdf119fc/stream/ws_client.py#L237

stream/ws_client.py
roycaihw: @PaulFurtado The statement about "file descriptor number under 1024" sounds reasonable. I didn't know that. I'm not against adding a separate `update_with_no_timeout` method if that solves the problem with `select`, as long as we don't break existing users who depend on the timeout interface

stream/ws_client.py
TomasTomecek: I am going to close this PR since this seems to be the solution to the problem.

Let me know if it's worth to document it (and I can send a PR).

Issue #142, opened 2019-07-03
OIDC auth behaivor differs from kubectl


mogaika:
I trying to use kubeconfig that works with kubectl but causes problems in python client
```
auth-provider:
  name: oidc
  config:
    client-id: <CENSORED>
    id-token: <CENSORED>
    idp-certificate-authority-data: <CENSORED>
    idp-issuer-url: <CENSORED>
    refresh-token: <CENSORED>
```
Code: ```config.load_kube_config('testkube.yml')```
Error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mogaika/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 549, in load_kube_config
    loader.load_and_set(config)
  File "/home/mogaika/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 430, in load_and_set
    self._load_authentication()
  File "/home/mogaika/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 194, in _load_authentication
    if self._load_auth_provider_token():
  File "/home/mogaika/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 213, in _load_auth_provider_token
    return self._load_oid_token(provider)
  File "/home/mogaika/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 290, in _load_oid_token
    self._refresh_oidc(provider)
  File "/home/mogaika/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py", line 351, in _refresh_oidc
    verify=config.ssl_ca_cert if config.verify_ssl else None
  File "/home/mogaika/.local/lib/python3.6/site-packages/requests_oauthlib/oauth2_session.py", line 363, in refresh_token
    timeout=timeout, headers=headers, verify=verify, withhold_token=True, proxies=proxies)
  File "/home/mogaika/.local/lib/python3.6/site-packages/requests/sessions.py", line 581, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/mogaika/.local/lib/python3.6/site-packages/requests_oauthlib/oauth2_session.py", line 425, in request
    headers=headers, data=data, **kwargs)
  File "/home/mogaika/.local/lib/python3.6/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
[...]

PR #144, opened 2019-07-10
Improve similarity with kubectl in handling of oidc kubeconfigs

mogaika:
- allow 'client-secret' to be empty
- fix 'verify' parameter of refresh_token call

Fixes: #142

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -361,13 +361,14 @@ class KubeConfigLoader(object):
             return
 
         response = json.loads(response.data)
+        client_secret = provider['config'].safe_get('client-secret')
 
         request = OAuth2Session(
             client_id=provider['config']['client-id'],
             token=provider['config']['refresh-token'],
             auto_refresh_kwargs={
                 'client_id': provider['config']['client-id'],
-                'client_secret': provider['config']['client-secret']
+                'client_secret': client_secret
             },
             auto_refresh_url=response['token_endpoint']
         )
@@ -377,7 +378,7 @@ class KubeConfigLoader(object):
                 token_url=response['token_endpoint'],
                 refresh_token=provider['config']['refresh-token'],
                 auth=(provider['config']['client-id'],
-                      provider['config']['client-secret']),
+                      client_secret),
                 verify=config.ssl_ca_cert if config.verify_ssl else None
             )
         except oauthlib.oauth2.rfc6749.errors.InvalidClientIdError:


yliaog:
i think you mean "Improve similarity with kubectl" instead of "Improve similarity with kubelet"

config/kube_config.py
```diff
@@ -377,8 +378,8 @@ def _refresh_oidc(self, provider):
                 token_url=response['token_endpoint'],
                 refresh_token=provider['config']['refresh-token'],
                 auth=(provider['config']['client-id'],
-                      provider['config']['client-secret']),
-                verify=config.ssl_ca_cert if config.verify_ssl else None
+                      client_secret),
+                verify=config.ssl_ca_cert if config.verify_ssl else False
```
roycaihw: As you pointed out in the issue, `verify` can be [either a boolean or a string](https://github.com/kennethreitz/requests/blob/ea9436a5d6a5934a906e91202349aabf9af75d15/requests/sessions.py#L499-L501). When `verify` is `None`, the requests Session tries to [look for requests environment configuration](https://github.com/kennethreitz/requests/blob/eaab47f033cae3bd443abb6707dfc30ae3482c36/requests/sessions.py#L704-L708). I wonder what's the implication of changing [`verify`](https://github.com/kennethreitz/requests/blob/eaab47f033cae3bd443abb6707dfc30ae3482c36/requests/sessions.py#L499-L501) to `False` here? More importantly, what's the expected behavior when `idp-certificate-authority-data` is empty in kubeconfig (and `verify_ssl` being False as a result)

roycaihw: @yliaog Do you have more context on this as you looked at https://github.com/kubernetes-client/python-base/pull/69?

yliaog: i don't have more context. please add a unit test that fails before the PR, and works after the PR.

scottilee:
@mogaika could you provide an update to the comment above?

config/kube_config.py
mogaika: My logic was based on the fact that kubectl ignores cert verification problems (like I had in #142). And since request method description notes that 'verify' parameter can be either bool or string, I thought that original code planned to use False instead of None, because its strange to explicitly point that 'verify' parameter must be None since it's None by default. I remove the parameter change from this commit since it's another unclear subject to discuss (if kubectl ignores verification problems is problem of kubectl or no, and does python client mimics this strange behaviour)

config/kube_config.py
```diff
@@ -361,13 +361,14 @@ def _refresh_oidc(self, provider):
             return
 
         response = json.loads(response.data)
+        client_secret = provider['config'].safe_get('client-secret') or ''
```
roycaihw: Should [client_secret](https://github.com/requests/requests-oauthlib/blob/36f24b494a146c60c47bfcdf5a0ae5fbc9a608b9/requests_oauthlib/oauth2_session.py#L229-L231) be empty string or None? Could you show what kubectl sends?

pshchelo: I've stumbled on the same issue, and while poking around, `None` is what works

mogaika:
I currently have no option to test real traffic but looking on kubectl client go code I confirm that @roycaihw and @pshchelo right

roycaihw:
/lgtm
/approve

allowing client-secret to be empty looks good. Please fix the `commit missing GitHub user` issue

z0r0:
Can we re-open this one? I'm facing the same issue here.

PR #145, opened 2019-07-24
Generate docs in python3

fabianvf:
Same as this change: https://github.com/kubernetes-client/python/pull/889


fabianvf:
@micw523

PR #152, opened 2019-08-14
Refs. #151 -- detect binary payloads and send the correct opcode

### LINKED ISSUE #151: Python3 write_channel "Can't convert 'bytes' object to str implicitly"
I am using the rundeck kubernetes plugin to execute a script to a pod and I get the following error
```
15:53:46 | Traceback (most recent call last):
-- | --
15:53:46 | File "/home/rundeck/libext/cache/kubernetes-plugin-1.10.1-SNAPSHOT/pods-copy-file.py", line 69, in <module>
15:53:46 | main()
15:53:46 | File "/home/rundeck/libext/cache/kubernetes-plugin-1.10.1-SNAPSHOT/pods-copy-file.py", line 65, in main
15:53:46 | common.copy_file(name, namespace, container, source_file, destination_path, destination_file_name)
15:53:46 | File "/home/rundeck/libext/cache/kubernetes-plugin-1.10.1-SNAPSHOT/common.py", line 396, in copy_file
15:53:46 | resp.write_stdin(c)
15:53:46 | File "/usr/local/lib/python3.5/dist-packages/kubernetes/stream/ws_client.py", line 160, in write_stdin
15:53:46 | self.write_channel(STDIN_CHANNEL, data)
15:53:46 | File "/usr/local/lib/python3.5/dist-packages/kubernetes/stream/ws_client.py", line 114, in write_channel
15:53:46 | self.sock.send(chr(channel) + data)
15:53:46 | TypeError: Can't convert 'bytes' object to str implicitly
```

On this [comment](https://github.com/kubernetes-client/python/issues/476#issuecomment-375056804) there is a workaround - a change to [ws_client.py](https://github.com/kubernetes-client/python-base/blob/master/stream/ws_client.py)
```
def write_channel(self, channel, data):
    """Write data to a channel."""
    self.sock.send(bytes(chr(channel), 'utf-8') + data)
```

This worked for me, so maybe someone with more knowledge can see if it should be merged fixed this way, or there's something better

I am using 'Python 3.5.2' and I tried kubernetes 9.0.0 and 10.0.0
This issue does not happen with Python2.7

sergei-maertens:
On Python 2, strings are bytestrings either way. On Python 3, the
result of `chr(channel)` is `str`, while the data itself is
`bytes`. The channel prefix needs to be turned into a binary type,
and the websocket frame needs the correct opcode (binary vs. text).

See #151 for the bug report and related issues.

This is basically the same fix as #52 but with a different approach and passes the `opcode` argument to indicate a binary frame.

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -116,7 +116,16 @@ class WSClient:
 
     def write_channel(self, channel, data):
         """Write data to a channel."""
-        self.sock.send(chr(channel) + data)
+        # check if we're writing binary data or not
+        binary = six.PY3 and type(data) == six.binary_type
+        opcode = ABNF.OPCODE_BINARY if binary else ABNF.OPCODE_TEXT
+
+        channel_prefix = chr(channel)
+        if binary:
+            channel_prefix = six.binary_type(channel_prefix, "ascii")
+
+        payload = channel_prefix + data
+        self.sock.send(payload, opcode=opcode)
 
     def peek_stdout(self, timeout=0):
         """Same as peek_channel with channel=1."""


sergei-maertens:
I'm not sure what the best way to test this is, without a real, functioning underlying websocket. The only test in `ws_client_test.py` didn't really give any good pointers either.

sergei-maertens:
What is needed to progress this PR? Anything I can do?

stream/ws_client.py
```diff
@@ -116,7 +116,16 @@ def readline_channel(self, channel, timeout=None):
...
+        # check if we're writing binary data or not
+        binary = six.PY3 and type(data) == six.binary_type
+        opcode = ABNF.OPCODE_BINARY if binary else ABNF.OPCODE_TEXT
+
+        channel_prefix = chr(channel)
+        if binary:
+            channel_prefix = six.binary_type(channel_prefix, "ascii")
+
+        payload = channel_prefix + data
+        self.sock.send(payload, opcode=opcode)
```
roycaihw: Could you add a test case [here](https://github.com/kubernetes-client/python/blob/ceaf188498bd43577b1bec72cc4e839be2379632/kubernetes/e2e_test/test_client.py#L38-L119) that writes binary data?

Ark-kun:
It would be nice to have a working support for getting binary data from pods.

roycaihw:
> It would be nice to have a working support for getting binary data from pods.

@Ark-kun this change focus on writing binary data. Please open a separate issue if there is a problem with reading binary data.

roycaihw:
A testcase that demonstrates the fix would be nice. @sergei-maertens Are you still interested in this issue?

sergei-maertens:
Yes and no - from the business I don't have the time available to work on
this, and the whole process to get contributions merged or reviewed works
very demotivating. I get why the bot is set up the way it is, but it
doesn't make our contributions seem appreciated. It's a very lengthy and
demotivating process.

roycaihw:
I just realized that the e2e test I asked for in https://github.com/kubernetes-client/python-base/pull/152#discussion_r324944046 belongs to a different repo. I agree in this case keeping python and python-base separate causes a bad contributor experience. @sergei-maertens I'm sorry for the confusion. 

I'm still not comfortable getting this bug fix in without a test, but the test and the bug being in two repos means we cannot merge them in the same time. Given the behavior change is well scoped, and the original bytes behavior is broken. I will merge this change and open an issue to track adding a test.

Issue #156, opened 2019-08-28
kubernetes.stream with proxy


AyliD:
Seems like even though i have setup proxy to all of my k8s python api calls i cannot use stream in order to perform the call to connect_get_namespaced_pod_exec

```
c = client.Configuration()
c.proxy = proxy_url
client.Configuration.set_default(c)

stream(api.connect_get_namespaced_pod_exec, 'some_name', 'default', command="/bin/sh", stderr=True, stdin=False, stdout=True, tty=True)
```

the class 'WSClient' doesnt seem to take this into account:
Example for the cod e i expected:
```
ws.connect = ('wss://blabla.com', http_proxy_host='ip', http_proxy_port='8080')
```

AyliD:
The exception im getting prior to fix is:
```
    (data) = self.connect_get_namespaced_pod_exec_with_http_info(name, namespace, **kwargs)
  File "/data/users/ayeletrd/fbsource/fbcode/buck-out/dev/gen/express_wifi/partners/setup/standalone/partner_setup#link-tree/kubernetes/client/apis/core_v1_api.py", line 935, in connect_get_namespaced
_pod_exec_with_http_info
    collection_formats=collection_formats)
  File "/data/users/ayeletrd/fbsource/fbcode/buck-out/dev/gen/express_wifi/partners/setup/standalone/partner_setup#link-tree/kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/data/users/ayeletrd/fbsource/fbcode/buck-out/dev/gen/express_wifi/partners/setup/standalone/partner_setup#link-tree/kubernetes/client/api_client.py", line 168, in __call_api
    _request_timeout=_request_timeout)
  File "/data/users/ayeletrd/fbsource/fbcode/buck-out/dev/gen/express_wifi/partners/setup/standalone/partner_setup#link-tree/kubernetes/stream/stream.py", line 31, in _intercept_request_call
    return ws_client.websocket_call(config, *args, **kwargs)
  File "/data/users/ayeletrd/fbsource/fbcode/buck-out/dev/gen/express_wifi/partners/setup/standalone/partner_setup#link-tree/kubernetes/stream/ws_client.py", line 259, in websocket_call
    raise ApiException(status=0, reason=str(e))
kubernetes.client.rest.ApiException: (0)
```

AyliD:
For all other API calls (except the exec) proxy works great.
```
[...]

PR #157, opened 2019-08-29
Update ws_client.py to support proxy

AyliD:
https://github.com/kubernetes-client/python-base/issues/156


roycaihw:
/lgtm
/approve

Thanks for fixing this!

PR #158, opened 2019-09-06
Check is not None in safe_get

qlemaire22:
Hi, 

I would add a check to see if `self.value` is not None in `safe_get`.

Otherwise it can create a `TypeError: argument of type 'NoneType' is not iterable` which I faced when working on Apache Airflow. I believe that I would be easier to debug.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -140,7 +140,11 @@ class KubeConfigLoader(object):
                  config_base_path="",
                  config_persister=None):
 
-        if isinstance(config_dict, ConfigNode):
+        if config_dict is None:
+            raise ConfigException(
+                'Invalid kube-config. '
+                'Expected config_dict to not be None.')
+        elif isinstance(config_dict, ConfigNode):
             self._config = config_dict
         else:
             self._config = ConfigNode('kube-config', config_dict)
@@ -612,6 +616,11 @@ def _get_kube_config_loader_for_yaml_file(
     kcfg = KubeConfigMerger(filename)
     if persist_config and 'config_persister' not in kwargs:
         kwargs['config_persister'] = kcfg.save_changes()
+
+    if kcfg.config is None:
+        raise ConfigException(
+            'Invalid kube-config file. '
+            'No configuration found.')
 
     return KubeConfigLoader(
         config_dict=kcfg.config,


qlemaire22:
I signed it!

config/kube_config.py
```diff
@@ -494,7 +494,7 @@ def __len__(self):
 
     def safe_get(self, key):
         if (isinstance(self.value, list) and isinstance(key, int) or
-                key in self.value):
+                self.value is not None and key in self.value):
```
roycaihw: Please add a test for this. Do you have a kubeconfig yaml that reproduces the error?

`self.value` should be either a list or a dict. I think we should restrict ConfigNade with `None` value early if we have such a case.

qlemaire22: You are right. In my case, it happened when calling _get_kube_config_loader_for_yaml_file  with no kubeconfig available. Should I raise an error somewhere like in KubeConfigMerger, KubeConfigLoader or ConfigNode to warn about empty kube_config?

roycaihw: could you add checks in both `_get_kube_config_loader_for_yaml_file` and `KubeConfigLoader`?

In `_get_kube_config_loader_for_yaml_file` an error should be raised if `kcfg.config` is None. This is likely be caused by invalid kubeconfig file path. This error restores the old [behavior](https://github.com/kubernetes-client/python-base/blob/4750aa9d3691cd0652654b56f54fb6897001a4a7/config/kube_config.py#L470-L475) we had

In `KubeConfigLoader` an error should be raised if the input `config_dict` is None. This makes the error more obvious if someone (unlikely) uses the `KubeConfigLoader` directly with a None `config_dict` object

qlemaire22: I did add the two checks. Please, tell me if I should put more specific error messages.

config/kube_config.py
```diff
@@ -140,7 +140,11 @@ def __init__(self, config_dict, active_context=None,
                  config_base_path="",
                  config_persister=None):
 
-        if isinstance(config_dict, ConfigNode):
+        if config_dict is None:
+            raise ConfigException(
+                'Invalid kube-config file. '
```
roycaihw: `config_dict` here doesn't necessarily come from a kubeconfig file. Change the error message to `Invalid kube config. Expected config_dict to not be None.`

qlemaire22: Right, I changed the error message.

PR #160, opened 2019-09-17
Add property returncode to WSClient

oz123:
This will be familiar for Python users as subprocess.Popen has
the same attribute. Also, the behavior is such the returncode returns
a None value if the process was not run yet.

Other than that, when the process exists with a numerical code this will
be the value of the return code. If the command executed successfully
the return value will be 0.

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -14,14 +14,18 @@
 
 from kubernetes.client.rest import ApiException
 
-import select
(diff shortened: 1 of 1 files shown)


roycaihw:
/lgtm
/approve

Thanks for having it e2e tested in https://github.com/kubernetes-client/python/pull/956!

PR #161, opened 2019-09-18
Support false values in configuration file

ganchurin:
The purpose of this changes is to fix the issue: https://github.com/kubernetes-client/python/issues/954

### DIFF
--- a/config/kube_config_test.py
+++ b/config/kube_config_test.py
@@ -564,13 +564,14 @@ class TestKubeConfigLoader(BaseTestCase):
                     "server": TEST_SSL_HOST,
                     "certificate-authority-data":
                         TEST_CERTIFICATE_AUTH_BASE64,
+                    "insecure-skip-tls-verify": False,
                 }
             },
             {
                 "name": "no_ssl_verification",
                 "cluster": {
                     "server": TEST_SSL_HOST,
-                    "insecure-skip-tls-verify": "true",
+                    "insecure-skip-tls-verify": True,
                 }
             },
         ],
@@ -1076,7 +1077,8 @@ class TestKubeConfigLoader(BaseTestCase):
             token=BEARER_TOKEN_FORMAT % TEST_DATA_BASE64,
             cert_file=self._create_temp_file(TEST_CLIENT_CERT),
             key_file=self._create_temp_file(TEST_CLIENT_KEY),
-            ssl_ca_cert=self._create_temp_file(TEST_CERTIFICATE_AUTH)
+            ssl_ca_cert=self._create_temp_file(TEST_CERTIFICATE_AUTH),
+            verify_ssl=True
         )
         actual = FakeConfig()
         KubeConfigLoader(
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -499,7 +499,7 @@ class ConfigNode(object):
 
     def __getitem__(self, key):
         v = self.safe_get(key)
-        if not v:
+        if v is None:
             raise ConfigException(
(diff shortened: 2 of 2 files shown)


config/kube_config_test.py
```diff
@@ -564,6 +564,7 @@ class TestKubeConfigLoader(BaseTestCase):
                     "server": TEST_SSL_HOST,
                     "certificate-authority-data":
                         TEST_CERTIFICATE_AUTH_BASE64,
+                    "insecure-skip-tls-verify": False,
```
roycaihw: s/False/"false" in consistent with the test case below

ganchurin: @roycaihw Thanks for the suggestion! I've just tried to replace `False` with `"false"` and faced an issue - the unit test fails. I think the cause is in how "insecure-skip-tls-verify" property is handled in `kube_config.py` (https://github.com/kubernetes-client/python-base/blob/master/config/kube_config.py#L443). The code by the link casts a string to a boolean. Both `"true"` and `"false"` are casted to `True`. What do you think if I replace `"true"` with `True` in the older test?

roycaihw: `"true"` and `"false"` are the original values in kubeconfig yaml files (e.g. `insecure-skip-tls-verify: false`). The test cases are testing values stored in the `KubeConfigLoader` object. 

Could you verify what's the behavior in the [yaml loader](https://github.com/kubernetes-client/python-base/blob/af42f24d726dc4a8e5aaca63ee30ccae93d64f1b/config/kube_config.py#L580)? Does it load `"false"` into a boolean `False`, or does it keep the string value `"false"`?

ganchurin: I verified the behavior on the configuration file which has the following lines:
```
clusters:
- cluster:
    insecure-skip-tls-verify: false
```

The following code snippet was used for the verification:
```
import yaml

with open(<PATH_TO_CONFIG>) as f:
    config = yaml.safe_load(f)
    print config (**breakpoint was set here**)
```

Debug showed the following value for the property:
`'insecure-skip-tls-verify' (4523129968) = {bool} False`

roycaihw: Thanks. Let's use `True` and `False` in the test cases then

roycaihw:
/lgtm
/approve

Thanks @ganchurin!

ganchurin:
@roycaihw Could you please help me a bit? Some tests have failed after my update to the branch. I'm not sure that they relate to the scope of work. And I can't restart test run. Thanks in advance!

micw523:
/ok-to-test
???? What’s happening here? I didn’t remember we required ok to test

micw523:
The Travis CI is showing a ton of 404 and other errors. Try looking at the log - it is available.

ganchurin:
@micw523 Thank you! I've taken a look.

I've noticed that the same tests fail in the latest build of "kubernetes-client/python" repository: https://travis-ci.org/kubernetes-client/python/builds/587650221

As for my pull request, previously the pull request had passed successfully. The single difference between successful and failed attempts is one line in unit tests that works fine for me locally. Thus I tend thinking that the issue is in "kubernetes-client/python" repository.

micw523:
The most recent build in k8s-python passed; I'm restarting your build to see if the test fail is a flake.

ganchurin:
@micw523 Sorry if I'm wrong - I was looking for builds here: https://travis-ci.org/kubernetes-client/python/builds. And the latest build is red.

roycaihw:
Test failure tracked in https://github.com/kubernetes-client/python/issues/963. Let's fix the CI first and then merge this PR

PR #167, opened 2019-10-02
Migrate to pytest

oz123:
As discussed previously, this replaces nose test runner with pytest.
This way there is a uniform toolset for python and python-base.


oz123:
This is a fix for #167.

micw523:
/lgtm
This lgtm to me.
Although, nosetests is still being used in the main repo to provide access to the code coverage reports of the base repo. This could also be switched to pytest as well -https://github.com/codecov/example-python

PR #169, opened 2019-10-04
config: Fix persist_config flag and function calls

jfrabaute:
The persist_config flag was setting the kwargs['config_persister'] to
the result of the function kcfg.save_changes and not the function
kcfg.save_changes itself.

Once this is fixed, the second problem was that the function was called
with an argument when it's defined without argument so an exception was
raised.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -336,7 +336,7 @@ class KubeConfigLoader(object):
         provider.value['access-token'] = token_response['accessToken']
         provider.value['expires-on'] = token_response['expiresOn']
         if self._config_persister:
-            self._config_persister(self._config.value)
+            self._config_persister()
 
     def _load_gcp_token(self, provider):
         if (('config' not in provider) or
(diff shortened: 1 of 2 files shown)


yliaog:
please sign cla
please add a test to verify the fix

jfrabaute:
CLA signed

jfrabaute:
Hi @yliaog 

Regarding the test, from what I understand from the code, the problem is in the function "_get_kube_config_loader_for_yaml_file" (this is where the fix is).
Should I add a test for this function? I'm asking because as it has a "_" it might just be a private function that should not be tested. The problem is that I can't find another way to test it as calling, as calling other functions using this one does not allow me to check that _config_persister is set correctly in the KubeConfigLoader object.
So, is that ok to add a test for "_get_kube_config_loader_for_yaml_file"?

Thanks.

jfrabaute:
I've added 2 tests for the "_get_kube_config_loader_for_yaml_file" function.
Keep me posted if that's correct or not.

Thanks.

jfrabaute:
Hi

Any news about reviewing this PR?

Thanks.

roycaihw:
/kind bug
/lgtm
/approve

Thanks for sending the fix!

PR #170, opened 2019-10-08
Fix azure refresh token apiserver id

### LINKED ISSUE #135: Azure Refresh Token Unauthorised Error
I am using python kubernetes client to connect my Cluster which is hosted in Azure. I am able to connect my kubernetes env using my token from kubeconfig file using python client, whenever my token expires,  I got a refreshed token but that token does not seems to be right. If I use that refreshed token thru python client or as well as manually from UI (Kubernetes) I am getting an unauthorized error. 

Little dig in to that method  _refresh_azure_token in kube_config.py
I see a client id mentioned as "00000002-0000-0000-c000-000000000000" - I assume this might by replaced with my client id - Please confirm. 

Either way I still getting that unauthorized error. Can some one help here? or direction? Appreciated

fooka03:
Fixes #135 

Currently, the refresh token logic is hardcoded to use the general Microsoft Graph application id as the apiserver.  This fails if there's a mismatch in the kubeconfig, specifically in the audience in the access token being refreshed.  Instead, the new code will dynamically load this from the local kubeconfig as with the other request values and use that to submit the refresh call.

Additionally, this sets the ADAL API version to 1.0 to keep the behavior consistent with the kubernetes go client azure authentication flow.  (Namely, the string `spi:` being included for the audience)

I also updated the tests to use the Microsoft Graph app id as that was the prior behavior the tests were relying on.

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -249,12 +249,15 @@ class KubeConfigLoader(object):
         tenant = config['tenant-id']
         authority = 'https://login.microsoftonline.com/{}'.format(tenant)
         context = adal.AuthenticationContext(
-            authority, validate_authority=True,
+            authority, validate_authority=True, api_version='1.0'
         )
         refresh_token = config['refresh-token']
         client_id = config['client-id']
+        apiserver_id = config.get(
+            'apiserver-id',
+            '00000002-0000-0000-c000-000000000000')
         token_response = context.acquire_token_with_refresh_token(
-            refresh_token, client_id, '00000002-0000-0000-c000-000000000000')
+            refresh_token, client_id, apiserver_id)
 
         provider = self._user['auth-provider']['config']
         provider.value['access-token'] = token_response['accessToken']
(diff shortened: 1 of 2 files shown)


config/kube_config.py
```diff
@@ -249,12 +249,13 @@ def _refresh_azure_token(self, config):
         tenant = config['tenant-id']
         authority = 'https://login.microsoftonline.com/{}'.format(tenant)
         context = adal.AuthenticationContext(
-            authority, validate_authority=True,
+            authority, validate_authority=True, api_version='1.0'
         )
         refresh_token = config['refresh-token']
         client_id = config['client-id']
+        apiserver_id = config['apiserver-id']
```
yliaog: this would break the kubeconfig that does not have the field 'apiserver-id'

fooka03: The azure provider requires apiserver-id unless I'm missing something in the workflow.  First thing I tried was removing apiserver-id from my kubeconfig to try and rely on the hardcoded ID.  I got an error about apiserver-id being a required field when trying to run commands.

PR #171, opened 2019-10-18
Fixed Watch.unmarshal_event when data is not a JSON-serialized object

Ark-kun:
Log streaming using Watch was introduced in https://github.com/kubernetes-client/python-base/pull/93/, but the users faced errors when the log line was a single number or an empty string.
This PR fixes those issues.

Fixes https://github.com/kubernetes-client/python/issues/982
Fixes https://github.com/kubernetes-client/python/issues/983

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -88,6 +88,8 @@ class Watch(object):
             js = json.loads(data)
         except ValueError:
             return data
+        if not (isinstance(js, dict) and 'object' in js):
+            return data
         js['raw_object'] = js['object']
         if return_type:
             obj = SimpleNamespace(data=json.dumps(js['raw_object']))


jiahuif:
As discussed here https://github.com/kubernetes-client/python/issues/982#issuecomment-544780986

If the object has a field named "object", the fix would not work. We should not rely on one of the field to determine the type.

Ark-kun:
> If the object has a field named "object", the fix would not work. We should not rely on one of the field to determine the type.

Unfortunately, dictionaries are duck-typed by design.
The code in question insists on accessing the 'object' key.
Without the fix, the code is randomly broken in large number of cases.

Can you please suggest another way to fix the common path?

scottilee:
@jiahuif any update for @Ark-kun ?

jiahuif:
@scottilee See kubernetes-client/python#1032

kvalev:
> Log streaming using Watch was introduced in #93, but the users faced errors when the log line was a single number or an empty string.

Also when the log line represents an empty array  `[]`, like we just had.

PR #175, opened 2019-10-23
Add option to refresh gcp token when config is cmd-path

jfrabaute:
The python client does not deal with "gcloud" gke auth in the kube config file as it's based on the "cmd-path" auth.

This is adding the code to take care of this auth.
This should solve kubernetes-client/python#754

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -20,8 +20,10 @@ import json
 import logging
 import os
 import platform
+import subprocess
 import tempfile
 import time
+from collections import namedtuple
 
 import google.auth
 import google.auth.transport.requests
@@ -133,6 +135,46 @@ class FileOrData(object):
         return self._data
 
 
+class CommandTokenSource(object):
+    def __init__(self, cmd, args, tokenKey, expiryKey):
+        self._cmd = cmd
+        self._args = args
+        if not tokenKey:
+            self._tokenKey = '{.access_token}'
+        else:
+            self._tokenKey = tokenKey
+        if not expiryKey:
+            self._expiryKey = '{.token_expiry}'
+        else:
+            self._expiryKey = expiryKey
+
+    def token(self):
+        fullCmd = self._cmd + (" ") + " ".join(self._args)
+        process = subprocess.Popen(
+            [self._cmd] + self._args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True)
+        (stdout, stderr) = process.communicate()
+        exit_code = process.wait()
+        if exit_code != 0:
+            msg = 'cmd-path: process returned %d' % exit_code
+            msg += "\nCmd: %s" % fullCmd
+            stderr = stderr.strip()
+            if stderr:
+                msg += '\nStderr: %s' % stderr
+            raise ConfigException(msg)
+        try:
+            data = json.loads(stdout)
+        except ValueError as de:
+            raise ConfigException(
+                'exec: failed to decode process output: %s' % de)
+        A = namedtuple('A', ['token', 'expiry'])
+        return A(
+            token=data['credential']['access_token'],
+            expiry=parse_rfc3339(data['credential']['token_expiry']))
+
+
 class KubeConfigLoader(object):
 
     def __init__(self, config_dict, active_context=None,
@@ -156,7 +198,38 @@ class KubeConfigLoader(object):
         self._config_base_path = config_base_path
         self._config_persister = config_persister
 
+        def _refresh_credentials_with_cmd_path():
+            config = self._user['auth-provider']['config']
+            cmd = config['cmd-path']
+            if len(cmd) == 0:
+                raise ConfigException(
+                    'missing access token cmd '
+                    '(cmd-path is an empty string in your kubeconfig file)')
+            if 'scopes' in config and config['scopes'] != "":
+                raise ConfigException(
+                    'scopes can only be used '
+                    'when kubectl is using a gcp service account key')
+            args = []
+            if 'cmd-args' in config:
+                args = config['cmd-args'].split()
+            else:
+                fields = config['cmd-path'].split()
+                cmd = fields[0]
+                args = fields[1:]
+
+            commandTokenSource = CommandTokenSource(
+                cmd, args,
+                config.safe_get('token-key'),
+                config.safe_get('expiry-key'))
+            return commandTokenSource.token()
+
         def _refresh_credentials():
+            # Refresh credentials using cmd-path
+            if ('auth-provider' in self._user and
+                'config' in self._user['auth-provider'] and
+                    'cmd-path' in self._user['auth-provider']['config']):
+                return _refresh_credentials_with_cmd_path()
+
             credentials, project_id = google.auth.default(scopes=[
                 'https://www.googleapis.com/auth/cloud-platform',
                 'https://www.googleapis.com/auth/userinfo.email'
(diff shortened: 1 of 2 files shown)


jfrabaute:
Looks like the test is failing on python style code problems on unrelated files (not thouched in that review).
I fixed a few, but I have one that I don't know about:

/tmp/tmp.0CpklmniPV/python/kubernetes/utils/quantity.py:24:80: E501 line too long (91 > 79 characters)

micw523:
The CI is failing on the files you modified. Take a look at the diff output - or to save time, just use ```autopep8 --aggressive --aggressive --in-place kube_config.py```.

E501 is ignored in judging the success of the CI task.

jfrabaute:
> The CI is failing on the files you modified. Take a look at the diff output - or to save time, just use `autopep8 --aggressive --aggressive --in-place kube_config.py`.
> 

Oops, yes. Fixed, build seems to pass now.

jfrabaute:
Hi,

Any chance to have this reviewed?

config/kube_config.py
```diff
@@ -156,7 +199,38 @@ def __init__(self, config_dict, active_context=None,
...
+                    '(cmd-path is an empty string in your kubeconfig file)')
+            if 'scopes' in config and config['scopes'] != "":
+                raise ConfigException(
+                    'scopes can only be used '
+                    'when kubectl is using a gcp service account key')
+            args = []
+            if 'cmd-args' in config:
+                args = config['cmd-args'].split()
+            else:
+                fields = config['cmd-path'].split()
```
yliaog: this is assuming config['cmd-path'] exists, probably better to check to ensure that.

config/kube_config.py
```diff
@@ -133,6 +135,47 @@ def as_data(self):
         return self._data
 
 
+class CommandTokenSource(object):
+    def __init__(self, cmd, args, tokenKey, expiryKey):
+
```
yliaog: why the extra blank line here?

config/kube_config_test.py
```diff
@@ -727,7 +731,8 @@ class TestKubeConfigLoader(BaseTestCase):
                     "auth-provider": {
                         "config": {
                             "access-token": TEST_AZURE_TOKEN,
-                            "apiserver-id": "00000002-0000-0000-c000-000000000000",
+                            "apiserver-id": "00000002-0000-0000-c000-"
```
yliaog: what is the reason for this change?

config/kube_config_test.py
```diff
@@ -1439,7 +1444,8 @@ def test_list_kube_config_contexts(self):
             {'context': {'cluster': 'ssl', 'user': 'ssl'}, 'name': 'ssl'},
             {'context': {'cluster': 'default', 'user': 'simple_token'},
              'name': 'simple_token'},
-            {'context': {'cluster': 'default', 'user': 'expired_oidc'}, 'name': 'expired_oidc'}]
+            {'context': {'cluster': 'default', 'user': 'expired_oidc'},
+             'name': 'expired_oidc'}]
```
yliaog: please add a test case

config/kube_config.py
jfrabaute: It should be set, otherwise it would have failed before (line 204 at least).
And this function is also called when 'cmd-path' is set so it should be set.

config/kube_config.py
jfrabaute: Fixed (when PR will be updated)

config/kube_config_test.py
jfrabaute: The code is exactly the same, but I had to split the string into 2 lines because some styling tests were failing with a line too long.
It's probably not related to my change, and it might be a styling check which has been added after this file was in the repo (not sure).
Maybe it should be a preliminary separate commit, not sure.

config/kube_config_test.py
jfrabaute: Working on it.

jfrabaute: Added tests and updated the PR.

config/kube_config_test.py
```diff
@@ -1394,6 +1501,7 @@ class TestKubeConfigMerger(BaseTestCase):
             {
                 "name": "expired_oidc",
                 "user": {
+                    "name": "gcp",
```
yliaog: why adding it? it does not relate to your code change, right?

config/kube_config_test.py
```diff
@@ -1279,6 +1346,46 @@ def test_user_exec_auth(self, mock):
...
+        token = "dummy"
+        return_value = A(token, parse_rfc3339(datetime.datetime.now()))
+        CommandTokenSource.token = mock.Mock(return_value=return_value)
+        expected = FakeConfig(api_key={
+                              "authorization": BEARER_TOKEN_FORMAT % token})
+        actual = FakeConfig()
+        KubeConfigLoader(
+            config_dict=self.TEST_KUBE_CONFIG,
+            active_context="contexttestcmdpath").load_and_set(actual)
+        expected.get_api_key_with_prefix = actual.get_api_key_with_prefix
```
yliaog: why? could you assert expected.api_key instead?

config/kube_config_test.py
```diff
@@ -1279,6 +1346,46 @@ def test_user_exec_auth(self, mock):
...
+
+    def test_user_cmd_path_empty(self):
+        A = namedtuple('A', ['token', 'expiry'])
+        token = "dummy"
+        return_value = A(token, parse_rfc3339(datetime.datetime.now()))
+        CommandTokenSource.token = mock.Mock(return_value=return_value)
+        expected = FakeConfig(api_key={
+                              "authorization": BEARER_TOKEN_FORMAT % token})
+        actual = FakeConfig()
+        with self.assertRaises(ConfigException):
```
yliaog: better to also check the raised exception is the correct one

config/kube_config_test.py
```diff
@@ -1279,6 +1346,46 @@ def test_user_exec_auth(self, mock):
...
+
+    def test_user_cmd_path_with_scope(self):
+        A = namedtuple('A', ['token', 'expiry'])
+        token = "dummy"
+        return_value = A(token, parse_rfc3339(datetime.datetime.now()))
+        CommandTokenSource.token = mock.Mock(return_value=return_value)
+        expected = FakeConfig(api_key={
+                              "authorization": BEARER_TOKEN_FORMAT % token})
+        actual = FakeConfig()
+        with self.assertRaises(ConfigException):
```
yliaog: better to also check the raised exception is the correct one

config/kube_config_test.py
jfrabaute: Ooops, my bad, fixed.

config/kube_config_test.py
jfrabaute: Done.

config/kube_config_test.py
jfrabaute: Done.

config/kube_config_test.py
jfrabaute: This is brekaing the "self.assertEqual" on the next line, because it looks like this attribute is transient and then it should not be part of the assertEqual, but right now this is how it's done, so I set this field as well, so the assert succeeds, but as this field is not important for the test, I just set it.
Is there a nicer way (I agree that it's ugly).

yliaog: what about clear get_api_key_with_prefix in actual?

jfrabaute: Done, Updated PR

jfrabaute:
Thanks @yliaog for the reviews.
I'll work on the other PR and add tests now: https://github.com/kubernetes-client/python-base/pull/169

palnabarun:
Note to self: We should add documentation for the functionality implemented here.

/cc

cc: @roycaihw

PR #176, opened 2019-11-11
Refresh OIDC tokens regularly

thoughteer:
Similar to GCP, OIDC tokens should be refreshed on regular basis. Currently, the OIDC token is refreshed only once, upon configuration loading. This causes `401 Unauthorized` errors when the refreshed token expires.


yliaog:
looks good, thanks. please add a test.

PR #178, opened 2019-11-27
stream+ws_client: Optional capture-all

rawler:
This PR is a proposal to fix https://github.com/kubernetes-client/python/issues/1013.

The fix is two-fold:
1) Improve the way WSClient._all is appended to, to avoid exponential slow-down.
2) Allow optional disabling of the "capture-all" behavior, to remove memory-usage altogether

### DIFF
--- a/stream/stream.py
+++ b/stream/stream.py
@@ -16,7 +16,8 @@ from . import ws_client
 
 
 def stream(func, *args, **kwargs):
-    """Stream given API call using websocket"""
+    """Stream given API call using websocket.
+    Extra kwarg: capture-all=True - captures all stdout+stderr for use with WSClient.read_all()"""
 
     def _intercept_request_call(*args, **kwargs):
         # old generated code's api client has config. new ones has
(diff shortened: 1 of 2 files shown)


rawler:
Side-note; As a user of the stream-API, the usage is not immediately clear, and it's also a bit inflexible. (Stdout is not necessarily characters?) I think this has room for some breaking API-improvements;
 - Adding clarity to what kwargs and I can and cannot specify - `stream()` kwargs should be explicit and documented
 - Supporting binary stdin/out. I had to work around this by use of base64. The data received seems to be bytes, the utf8-assumption is done in https://github.com/kubernetes-client/python-base/blob/a2d1024524de78b62e5f9aa72d34cb9ea9de2b97/stream/ws_client.py#L190
 - Buffering `_all` could potentially be a composed opt-in, rather than a default-integrated behavior.

I made a short attempt to add explicit `**kwargs` (which could be non-breaking), but it snowballed into scope-creep, and I realized I don't know the code well enough to feel comfortable making such changes.

rawler:
I signed it

roycaihw:
/lgtm
/approve

Thanks for the improvement!

PR #184, opened 2020-02-21
Fix apiserver_id 'get' method

### LINKED ISSUE #179: AttributeError: 'ConfigNode' object has no attribute 'get'
Using latest pre-release version ` kubernetes-11.0.0b2` with `Python 3.6.5` on `Windows 10` against an Azuze Kubernetes Service cluster, configured using Azure AD for authentication:

I receive the error:

```
  File "check_neptune2.py", line 4, in <module>
    config.load_kube_config()
  File "C:\programs\Python36\lib\site-packages\kubernetes\config\kube_config.py", line 667, in load_kube_config
    loader.load_and_set(config)
  File "C:\programs\Python36\lib\site-packages\kubernetes\config\kube_config.py", line 474, in load_and_set
    self._load_authentication()
  File "C:\programs\Python36\lib\site-packages\kubernetes\config\kube_config.py", line 205, in _load_authentication
    if self._load_auth_provider_token():
  File "C:\programs\Python36\lib\site-packages\kubernetes\config\kube_config.py", line 222, in _load_auth_provider_token
    return self._load_azure_token(provider)
  File "C:\programs\Python36\lib\site-packages\kubernetes\config\kube_config.py", line 241, in _load_azure_token
    self._refresh_azure_token(provider['config'])
  File "C:\programs\Python36\lib\site-packages\kubernetes\config\kube_config.py", line 256, in _refresh_azure_token
    apiserver_id = config.get(
AttributeError: 'ConfigNode' object has no attribute 'get'
```

fooka03:
Fixes #179 

Not sure how I missed this in testing locally but ConfigNode definitely does not have a .get method.  This change instead relies on __getitem__ and it throwing a ConfigException to fall back to the default value.

This commit also prunes some excess tests that are already covered by other cases.

### DIFF
--- a/config/kube_config_test.py
+++ b/config/kube_config_test.py
@@ -461,20 +461,6 @@ class TestKubeConfigLoader(BaseTestCase):
                 }
             },
             {
-                "name": "azure_no_apiserver",
-                "context": {
-                    "cluster": "default",
-                    "user": "azure_no_apiserver"
-                }
-            },
-            {
-                "name": "azure_bad_apiserver",
-                "context": {
-                    "cluster": "default",
-                    "user": "azure_bad_apiserver"
-                }
-            },
-            {
                 "name": "expired_oidc",
                 "context": {
                     "cluster": "default",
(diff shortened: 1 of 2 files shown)


config/kube_config_test.py
```diff
@@ -1161,22 +1114,6 @@ def test_azure_with_expired_int_error(self):
...
 
-    def test_azure_with_no_apiserver(self):
-        loader = KubeConfigLoader(
-            config_dict=self.TEST_KUBE_CONFIG,
-            active_context="azure_no_apiserver",
-        )
-        provider = loader._user['auth-provider']
-        self.assertTrue(loader._azure_is_expired(provider))
-
-    def test_azure_with_bad_apiserver(self):
```
roycaihw: The code fix looks good to me. Could you point out which other test cases cover these pruned cases?

fooka03: These two tests are calling the same function as `test_azure_with_expired_num` with the same arguments.  The apiserver values here aren't actually used in `_azure_is_expired` and `_refresh_azure_token` is not currently covered by tests (which is why this slipped through in the first place, and the test coverage didn't really change)
https://github.com/kubernetes-client/python-base/pull/184/files#diff-3123bf09ac4950ea8bf8de46fb38eb83L1132

PR #185, opened 2020-02-21
watch stream: don't unmarshal responses when streaming log

roycaihw:
We were doing a best-effort unmarshalling for both pod log and regular watches, but we really shouldn't try unmarshalling pod logs into watch events. The existing code would unmarshal a pod log that happens to be a JSON blob and contains a field "object". 

/hold 
We need an e2e test for basic watch functionality in the main repo first. 

/cc @Ark-kun @palnabarun

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -84,12 +84,7 @@ class Watch(object):
             return 'watch'
 
     def unmarshal_event(self, data, return_type):
-        try:
-            js = json.loads(data)
-        except ValueError:
-            return data
-        if not (isinstance(js, dict) and 'object' in js):
-            return data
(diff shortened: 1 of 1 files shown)


roycaihw:
adding a test in main repo: https://github.com/kubernetes-client/python/pull/1087

roycaihw:
```
kubernetes/e2e_test/test_watch.py::TestClient::test_watch_configmaps Running test against : https://10.20.255.31:8443
PASSED
```

/hold cancel

This change is ready for review. Please take a look

palnabarun:
Thank you for walking me through the changes and the background personally. :)

/lgtm

PR #187, opened 2020-03-13
Prevent 503s from killing the client during discovery

fabianvf:
Sometimes aggregated API servers are down, which results in a 503 during discovery. Rather than raising an exception during discovery, we'll instead ignore the error, and an error won't be raised until the user attempts to access that API. This allows users to ignore downed APIs that they don't use.

### DIFF
--- a/dynamic/discovery.py
+++ b/dynamic/discovery.py
@@ -23,7 +23,7 @@ from abc import abstractmethod, abstractproperty
 from urllib3.exceptions import ProtocolError, MaxRetryError
 
 from kubernetes import __version__
-from .exceptions import NotFoundError, ResourceNotFoundError, ResourceNotUniqueError, ApiException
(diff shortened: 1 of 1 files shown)


dynamic/discovery.py
```diff
@@ -155,7 +155,10 @@ def get_resources_for_api_version(self, prefix, group, version, preferred):
         subresources = {}
 
         path = '/'.join(filter(None, [prefix, group, version]))
-        resources_response = self.client.request('GET', path).resources or []
+        try:
+            resources_response = self.client.request('GET', path).resources or []
+        except ServiceUnavailableError:
+            resources_response = []
```
palnabarun: I think when the exception is caught we should probably log a warning. What do you think?

```
logger.warn('some helpful text')
```

fabianvf: makes sense, though this module is lacking logging in general. I can add it here and then maybe do a followup PR to add it to rest?

palnabarun: Sounds good to me. :+1:

PR #191, opened 2020-04-13
set expiration on token of incluster config and reload if expires

zshihang:
be default, this will set an expiration period of 1 minutes on service account token. this PR is solving kubernetes-client/python/issues/616. as we are graduating the Beta feature Projected Service Account Volume to GA, we expect clients to reload token before it expires.

### DIFF
--- a/config/incluster_config.py
+++ b/config/incluster_config.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import datetime
 
 from kubernetes.client import Configuration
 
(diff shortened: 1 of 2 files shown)


config/incluster_config.py
```diff
@@ -76,19 +75,37 @@ def _load_config(self):
...
         Configuration.set_default(configuration)
+        if not refresh_token:
+            return
+        def wrap(f):
+            in_cluster_config = self
+            def wrapped(self, identifier):
+                if identifier == 'authorization' and identifier in self.api_key and in_cluster_config.token_expires_at <= datetime.datetime.now():
+                    in_cluster_config._read_token_file()
+                    self.api_key[identifier] = "bearer " + in_cluster_config.token
+                return f(self, identifier)
```
roycaihw, line 92: nit: I'm a little concerned what if someone calls `load_and_set` more than once in the same program. Will we end up with nested if conditions?

zshihang: good point. i assumed `load_and_set` will only be called once. maybe enforce this restriction in a separate PR

PR #193, opened 2020-05-12
allow incluster to accept pass-in config

zshihang:
fixes the e2e tests in kubernetes-client/python/pull/1161.

in_cluster_config with refreshing is mutating the global Configuration class. since all the tests are running in a single process, e2e tests running after the unit tests will use a modified Configuration.

### DIFF
--- a/config/incluster_config.py
+++ b/config/incluster_config.py
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import datetime
+import os
 
 from kubernetes.client import Configuration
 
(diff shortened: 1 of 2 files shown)


zshihang:
travis has passed: https://travis-ci.org/github/kubernetes-client/python-base/builds/686358803

config/incluster_config.py
```diff
@@ -35,30 +35,33 @@ def _join_host_port(host, port):
 
 
 class InClusterConfigLoader(object):
-
-    def __init__(self, token_filename,
-                 cert_filename, environ=os.environ):
+    def __init__(self,
+                 token_filename,
+                 cert_filename,
+                 try_refresh_token,
```
roycaihw, line 41: This is a breaking change. Please add a default value

config/incluster_config.py
```diff
@@ -35,30 +35,33 @@ def _join_host_port(host, port):
...
+                 environ=os.environ):
         self._token_filename = token_filename
         self._cert_filename = cert_filename
         self._environ = environ
+        self._try_refresh_token = try_refresh_token
         self._token_refresh_period = datetime.timedelta(minutes=1)
 
-    def load_and_set(self, refresh_token=True):
+    def load_and_set(self, client_configuration):
```
roycaihw, line 49: Can we add a separate method and keep the existing one backwards-compatible?

config/incluster_config.py
zshihang: done

config/incluster_config.py
zshihang: added default value

config/incluster_config.py
```diff
@@ -35,30 +35,39 @@ def _join_host_port(host, port):
...
         self._cert_filename = cert_filename
         self._environ = environ
+        self._try_refresh_token = try_refresh_token
         self._token_refresh_period = datetime.timedelta(minutes=1)
 
-    def load_and_set(self, refresh_token=True):
+    def load_and_set(self, client_configuration=None):
+        try_set_default = False
+        if client_configuration is None:
+            config = type.__call__(Configuration)
```
roycaihw, line 52: s/config/client_configuration?

zshihang: good catch

PR #195, opened 2020-05-14
Adding ability to set kube config from a dict.

vishnu667:
Added a function called `load_kube_config_from_dict()`, adding the ability to set the `~/.kube/config` file from a dict.

Required where kubeconfig is stored externally and saving to disk is not an option.

#### Usage 
```
import yaml
import kubernetes
from os.path import expanduser
config_dict = yaml.load(expanduser("~") + "/.kube/config")
kubernetes.config.load_kube_config_from_dict(config_dict)
```

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -706,13 +706,34 @@ def _get_kube_config_loader_for_yaml_file(
         config_base_path=None,
         **kwargs)
 
+def _get_kube_config_loader(
+        filename=None,config_dict=None, persist_config=False, **kwargs):
+
+    if (config_dict is None):
+        kcfg = KubeConfigMerger(filename)
+        if persist_config and 'config_persister' not in kwargs:
+            kwargs['config_persister'] = kcfg.save_changes
+
+        if kcfg.config is None:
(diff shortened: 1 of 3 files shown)


yliaog:
please add tests for the change to kube_config_test.py

vishnu667:
@yliaog I've added Some test cases, they are similar to the `load_kube_config` test cases.
Instead of passing the yaml files I've passed them as a dictionary.

config/kube_config.py
```diff
@@ -745,6 +757,36 @@ def load_kube_config(config_file=None, context=None,
     else:
         loader.load_and_set(client_configuration)
 
+def load_kube_config_from_dict(config_dict, context=None,
+                     client_configuration=None,
+                     persist_config=True):
+    """Loads authentication and cluster information from kube-config file
```
yliaog, line 763: not from kube-config file, instead, from config_dict

vishnu667: Made the change @yliaog. Is there anything else that I should change ?
Especially in the test cases.

Issue #183, opened 2020-02-19
kube_config.py is unable to merge configs when one of configs has removed all contexts


matihost:
```
  File \"/home/myuser/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py\", line 664, in load_kube_config
    persist_config=persist_config)
  File \"/home/myuser/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py\", line 620, in _get_kube_config_loader_for_yaml_file
    kcfg = KubeConfigMerger(filename)
  File \"/home/myuser/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py\", line 574, in __init__
    self.load_config(path)
  File \"/home/myuser/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py\", line 592, in load_config
    self._merge(item, config.get(item, {}), path)
  File \"/home/myuser/.local/lib/python3.6/site-packages/kubernetes/config/kube_config.py\", line 597, in _merge
    for new_item in add_cfg:
TypeError: 'NoneType' object is not iterable
```

when:
```bash
KUBECONFIG=/home/myuser/.kube/config:/home/myuser/.kube/microk8s.config
```
and for example one of them defines null contexts:
```yaml
contexts: null
```
It happens when 
```
kubectl config delete-context microk8s 
```
is invoked and there is no context left in kube config file.

Fix proposal add if:
```python
    def _merge(self, item, add_cfg, path):
        if add_cfg is not None:
          for new_item in add_cfg:
```
around
https://github.com/kubernetes-client/python-base/blob/a2d1024524de78b62e5f9aa72d34cb9ea9de2b97/config/kube_config.py#L596

weininghu1012:
Hi, I also encounter this issue with error like: 

```
  File "/Users/weininghu/.pyenv/virtualenvs/kafka-topic-manager/3.7.8/lib/python3.7/site-packages/kubernetes/config/kube_config.py", line 672, in _merge
    for new_item in add_cfg:
TypeError: 'NoneType' object is not iterable
```

Is this issue being working on?

mohammedi-haroune:
I'm also getting this error
```
    def _merge(self, item, add_cfg, path):
>       for new_item in add_cfg:
E       TypeError: 'NoneType' object is not iterable

../../venv/lib/python3.7/site-packages/kubernetes/config/kube_config.py:672: TypeError
```

Issue #181, opened 2020-01-23
kube_config.py fails to load config if no context is set


micw:
I have an interactive application that let a user select a context (from the available contexts of a configuration). That works great but loading of the config fails if there's no active context set and no context is passed to the loader. In my case I cannot pass a context to the loader since I know contexts only after loading.

Solution: if no context is passed and no current-context is set, start without an active context.

https://github.com/kubernetes-client/python-base/blob/a2d1024524de78b62e5f9aa72d34cb9ea9de2b97/config/kube_config.py#L175

PR #198, opened 2020-06-02
fix cfg is none, load kube config error

ACXLM:
This PR fixed
https://github.com/kubernetes/kubernetes/pull/84503 changed the behaviour of kubectl config unset slightly that if no users left in kubeconfig after user run kubectl config unset users.<user>, the value of users field in kubeconfig would be set to null rather than [].

since users set to null,  kubernetes.config.kube_config.load_kube_config() can't work, this pr set a default empty list to users

Which issue(s) this PR fixes:

Try to fixes
https://github.com/kubernetes-client/python-base/issues/183
https://github.com/kubernetes-client/python-base/issues/181

### DIFF
--- a/config/kube_config_test.py
+++ b/config/kube_config_test.py
@@ -1342,7 +1342,7 @@ class TestKubeConfigLoader(BaseTestCase):
         actual = _get_kube_config_loader_for_yaml_file(config_file,
                                                        persist_config=True)
         self.assertTrue(callable(actual._config_persister))
-        self.assertEquals(actual._config_persister.__name__, "save_changes")
+        self.assertEqual(actual._config_persister.__name__, "save_changes")
 
 
 class TestKubernetesClientConfiguration(BaseTestCase):
@@ -1517,6 +1517,26 @@ class TestKubeConfigMerger(BaseTestCase):
             }
         ]
     }
+    TEST_KUBE_CONFIG_PART6 = {
+        "current-context": "no_user",
+        "contexts": [
+            {
+                "name": "no_user",
+                "context": {
+                    "cluster": "default"
+                }
+            },
+        ],
+        "clusters": [
+            {
+                "name": "default",
+                "cluster": {
+                    "server": TEST_HOST
+                }
+            },
+        ],
+        "users": None
+    }
 
     def _create_multi_config(self):
         files = []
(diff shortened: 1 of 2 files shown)


roycaihw:
Could you add a unit test for this behavior? Thanks

ACXLM:
> Could you add a unit test for this behavior? Thanks

since users set to None, load_config() run like the following error:

.....................config/kube_config_test.py:1345: DeprecationWarning: Please use assertEqual instead.
  self.assertEquals(actual._config_persister.__name__, "save_changes")
....................................EEE...
======================================================================
ERROR: test_list_kube_config_contexts (__main__.TestKubeConfigMerger)
----------------------------------------------------------------------
  File "config/kube_config_test.py", line 1545, in test_list_kube_config_contexts
    contexts, active_context = list_kube_config_contexts(
  File "/home/ac/Project/PythonWorksPaces/python-base/config/kube_config.py", line 714, in list_kube_config_contexts
    loader = _get_kube_config_loader_for_yaml_file(config_file)
  File "/home/ac/Project/PythonWorksPaces/python-base/config/kube_config.py", line 694, in _get_kube_config_loader_for_yaml_file
    kcfg = KubeConfigMerger(filename)
  File "/home/ac/Project/PythonWorksPaces/python-base/config/kube_config.py", line 650, in __init__
    self.load_config(path)
  File "/home/ac/Project/PythonWorksPaces/python-base/config/kube_config.py", line 667, in load_config
    self._merge(item, config.get(item, []), path)
  File "/home/ac/Project/PythonWorksPaces/python-base/config/kube_config.py", line 671, in _merge
    for new_item in add_cfg:
TypeError: 'NoneType' object is not iterable

this PR can fix this error

config/kube_config_test.py
```diff
@@ -1392,7 +1392,7 @@ class TestKubeConfigMerger(BaseTestCase):
                 }
             },
         ],
-        "users": []
+        "users": None
```
roycaihw, line 1395: Could you add a separate test case and keep the existing one unchanged? It's better to make sure kube_config also works with the old `kubectl config unset` behavior for old kubectl versions. 

LGTM otherwise

roycaihw:
/lgtm
/hold
LGTM once the pycodestyle failure is fixed: https://travis-ci.org/github/kubernetes-client/python-base/jobs/701111762

ACXLM:
> /lgtm
> /hold
> LGTM once the pycodestyle failure is fixed: https://travis-ci.org/github/kubernetes-client/python-base/jobs/701111762

fixed pycodestyle test failure

PR #199, opened 2020-06-20
resource version reset default when event object resource version inv…

DrAuYueng:
Watch an API resource and stream the result back via a generator, when resource_version specified,  may be cause "too old resource version" error, then result would be:
`{
    "type":"ERROR",
    "object":{
        "kind":"Status",
        "apiVersion":"v1",
        "metadata":{
        },
        "status":"Failure",
        "message":"too old resource version: 38081900 (64845263)",
        "reason":"Gone",
        "code":410
    }
}
`
In this case, the func unmarshal_event can not parse the right  resource version, actually is None,
the stream while loop continue, the next request will get ApiException: (500) error:
`HTTP response headers: HTTPHeaderDict({'Content-Length': '186', 'Content-Type': 'application/json', 'Date': 'Fri, 19 Jun 2020 06:21:06 GMT'})
HTTP response body: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"resourceVersion: Invalid value: \\"None\\": strconv.ParseUint: parsing \\"None\\": invalid syntax","code":500}\n'`

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -97,6 +97,10 @@ class Watch(object):
                   and 'resourceVersion' in js['object']['metadata']):
                 self.resource_version = js['object']['metadata'][
                     'resourceVersion']
+            if self.resource_version is None:
+                # In some case like 'too old resource version',
+                # the event type would be ERROR, resource version will be None
+                self.resource_version = 0
         return js
 
     def stream(self, func, *args, **kwargs):


DrAuYueng:
@yliaog

DrAuYueng:
kubernetes version：v1.14.6

watch/watch.py
```diff
@@ -97,6 +97,10 @@ def unmarshal_event(self, data, return_type):
                   and 'resourceVersion' in js['object']['metadata']):
                 self.resource_version = js['object']['metadata'][
                     'resourceVersion']
+            if self.resource_version is None:
+                # In some case like 'too old resource version',
+                # the event type would be ERROR, resource version will be None
+                self.resource_version = 0
```
roycaihw, line 103: Watching with `resourceVersion=0` [doesn't guarantee](https://github.com/kubernetes/kubernetes/blob/9cd906e9322f75ce48b47613dd9e025c2b06c143/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L362) the client to continue watching from the last seen state, and therefore may cause missing watch events. The right way to handle a `410 Gone` response is to perform a LIST request, and start the watch from the `resourceVersion` returned by that new list operation (see https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes for more details). 

This is part of the problems https://github.com/kubernetes-client/python/issues/868 was trying to address

DrAuYueng: If  catch event type of Error and continue like #133 ,will cause endless loop

DrAuYueng: > If catch event type of Error and continue like #133 ,will cause endless loop, current resourceVersion never change, always get Error of message:"too old resource version"

Make resourceVersion=0 make sure the watch stream still work even if the wrong init resource version set

roycaihw: > will cause endless loop, current resourceVersion never change, always get Error of message:"too old resource version"

I don't think that's what we want and it should not happen

> perform a LIST request, and start the watch from the resourceVersion returned by that new list operation

the resourceVersion returned by a LIST request will be the latest resourceVersion, it won't be stale

PR #202, opened 2020-07-16
Add old api_key to set auth attributes

roycaihw:
fixes https://github.com/kubernetes-client/python/pull/1188#issuecomment-659207213

Add "authorization" to api_key with a old token, so that Configuration returns the overwritten auth attributes. 

/assign @palnabarun


palnabarun:
/lgtm

Thank you @roycaihw :slightly_smiling_face:

PR #204, opened 2020-07-16
Fix a Python 2 compatibility issue

palnabarun:
PR #133 introduces the usage of `http` module for checking the status
    code for `GONE` HTTP status. However, this doesn't work in Python 2.7.
    
    This commit checks if the interpreter is Python 2 and imports the
    status code from `httplib` module instead and unifies the approach
    to the checks.


yliaog:
thanks for the fix

/lgtm
/approve

PR #206, opened 2020-07-31
Leader Election issue #434

Invictus17:
Leader Election issue #434
https://github.com/kubernetes-client/python/issues/434

### DIFF
--- a/leaderelection/README.md
+++ b/leaderelection/README.md
@@ -1,0 +1,18 @@
+## Leader Election Example  
+This example demonstrates how to use the leader election library.
+
+## Running
+Run the following command in multiple separate terminals preferably an odd number. 
+Each running process uses a unique identifier displayed when it starts to run.
+
+- When a program runs, if a lock object already exists with the specified name, 
+all candidates will start as followers. 
+- If a lock object does not exist with the specified name then whichever candidate 
+creates a lock object first will become the leader and the rest will be followers.
+- The user will be prompted about the status of the candidates and transitions. 
+
+### Command to run
+```python example.py```
+
+Now kill the existing leader. You will see from the terminal outputs that one of the
+ remaining running processes will be elected as the new leader.
--- a/leaderelection/example.py
+++ b/leaderelection/example.py
@@ -1,0 +1,54 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import uuid
+from kubernetes import client, config
+from leaderelection import leaderelection
+from leaderelection.resourcelock.configmaplock import ConfigMapLock
+from leaderelection import electionconfig
+
+
+# Authenticate using config file
+config.load_kube_config(config_file=r"")
+
+# Parameters required from the user
+
+# A unique identifier for this candidate
+candidate_id = uuid.uuid4()
+
+# Name of the lock object to be created
+lock_name = "examplepython"
+
+# Kubernetes namespace
+lock_namespace = "default"
+
+
+# The function that a user wants to run once a candidate is elected as a leader
+def example_func():
+    print("I am leader")
+
+
+# A user can choose not to provide any callbacks for what to do when a candidate fails to lead - onStoppedLeading()
+# In that case, a default callback function will be used
+
+# Create config
+config = electionconfig.Config(ConfigMapLock(lock_name, lock_namespace, candidate_id), lease_duration=17,
+                               renew_deadline=15, retry_period=5, onstarted_leading=example_func,
+                               onstopped_leading=None)
+
+# Enter leader election
+leaderelection.LeaderElection(config).run()
+
+# User can choose to do another round of election or simply exit
+print("Exited leader election")
--- a/leaderelection/leaderelection.py
+++ b/leaderelection/leaderelection.py
@@ -1,0 +1,191 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import datetime
+import sys
+import time
+import json
+import threading
+from .leaderelectionrecord import LeaderElectionRecord
+import logging
+# if condition to be removed when support for python2 will be removed
+if sys.version_info > (3, 0):
+    from http import HTTPStatus
+else:
+    import httplib
+logging.basicConfig(level=logging.INFO)
+
+"""
+This package implements leader election using an annotation in a Kubernetes object.
+The onstarted_leading function is run in a thread and when it returns, if it does 
+it might not be safe to run it again in a process.
+
+At first all candidates are considered followers. The one to create a lock or update
+an existing lock first becomes the leader and remains so until it keeps renewing its
+lease.
+"""
+
+
+class LeaderElection:
+    def __init__(self, election_config):
+        if election_config is None:
+            sys.exit("argument config not passed")
+
+        # Latest record observed in the created lock object
+        self.observed_record = None
+
+        # The configuration set for this candidate
+        self.election_config = election_config
+
+        # Latest update time of the lock
+        self.observed_time_milliseconds = 0
+
+    # Point of entry to Leader election
+    def run(self):
+        # Try to create/ acquire a lock
+        if self.acquire():
+            logging.info("{} successfully acquired lease".format(self.election_config.lock.identity))
+
+            # Start leading and call OnStartedLeading()
+            threading.daemon = True
+            threading.Thread(target=self.election_config.onstarted_leading).start()
+
+            self.renew_loop()
+
+            # Failed to update lease, run OnStoppedLeading callback
+            self.election_config.onstopped_leading()
+
+    def acquire(self):
+        # Follower
+        logging.info("{} is a follower".format(self.election_config.lock.identity))
+        retry_period = self.election_config.retry_period
+
+        while True:
+            succeeded = self.try_acquire_or_renew()
+
+            if succeeded:
+                return True
+
+            time.sleep(retry_period)
+
+    def renew_loop(self):
+        # Leader
+        logging.info("Leader has entered renew loop and will try to update lease continuously")
+
+        retry_period = self.election_config.retry_period
+        renew_deadline = self.election_config.renew_deadline * 1000
+
+        while True:
+            timeout = int(time.time() * 1000) + renew_deadline
+            succeeded = False
+
+            while int(time.time() * 1000) < timeout:
+                succeeded = self.try_acquire_or_renew()
+
+                if succeeded:
+                    break
+                time.sleep(retry_period)
+
+            if succeeded:
+                time.sleep(retry_period)
+                continue
+
+            # failed to renew, return
+            return
+
+    def try_acquire_or_renew(self):
+        now_timestamp = time.time()
+        now = datetime.datetime.fromtimestamp(now_timestamp)
+
+        # Check if lock is created
+        lock_status, old_election_record = self.election_config.lock.get(self.election_config.lock.name,
+                                                                        self.election_config.lock.namespace)
+
+        # create a default Election record for this candidate
+        leader_election_record = LeaderElectionRecord(self.election_config.lock.identity,
+                                                     str(self.election_config.lease_duration), str(now), str(now))
+
+        # A lock is not created with that name, try to create one
+        if not lock_status:
+            # To be removed when support for python2 will be removed
+            if sys.version_info > (3, 0):
+                if json.loads(old_election_record.body)['code'] != HTTPStatus.NOT_FOUND:
+                    logging.info("Error retrieving resource lock {} as {}".format(self.election_config.lock.name,
+                                                                                  old_election_record.reason))
+                    return False
+            else:
+                if json.loads(old_election_record.body)['code'] != httplib.NOT_FOUND:
+                    logging.info("Error retrieving resource lock {} as {}".format(self.election_config.lock.name,
+                                                                                  old_election_record.reason))
+                    return False
+
+            logging.info("{} is trying to create a lock".format(leader_election_record.holder_identity))
+            create_status = self.election_config.lock.create(name=self.election_config.lock.name,
+                                                             namespace=self.election_config.lock.namespace,
+                                                             election_record=leader_election_record)
+
+            if create_status is False:
+                logging.info("{} Failed to create lock".format(leader_election_record.holder_identity))
+                return False
+
+            self.observed_record = leader_election_record
+            self.observed_time_milliseconds = int(time.time() * 1000)
+            return True
+
+        # A lock exists with that name
+        # Validate old_election_record
+        if old_election_record is None:
+            # try to update lock with proper annotation and election record
+            return self.update_lock(leader_election_record)
+
+        if (old_election_record.holder_identity is None or old_election_record.lease_duration is None
+                or old_election_record.acquire_time is None or old_election_record.renew_time is None):
+            # try to update lock with proper annotation and election record
+            return self.update_lock(leader_election_record)
+
+        # Report transitions
+        if self.observed_record and self.observed_record.holder_identity != old_election_record.holder_identity:
+            logging.info("Leader has switched to {}".format(old_election_record.holder_identity))
+
+        if self.observed_record is None or old_election_record.__dict__ != self.observed_record.__dict__:
+            self.observed_record = old_election_record
+            self.observed_time_milliseconds = int(time.time() * 1000)
+
+        # If This candidate is not the leader and lease duration is yet to finish
+        if (self.election_config.lock.identity != self.observed_record.holder_identity
+                and self.observed_time_milliseconds + self.election_config.lease_duration * 1000 > int(now_timestamp * 1000)):
+            logging.info("yet to finish lease_duration, lease held by {} and has not expired".format(old_election_record.holder_identity))
+            return False
+
+        # If this candidate is the Leader
+        if self.election_config.lock.identity == self.observed_record.holder_identity:
+            # Leader updates renewTime, but keeps acquire_time unchanged
+            leader_election_record.acquire_time = self.observed_record.acquire_time
+
+        return self.update_lock(leader_election_record)
+
+    def update_lock(self, leader_election_record):
+        # Update object with latest election record
+        update_status = self.election_config.lock.update(self.election_config.lock.name,
+                                                         self.election_config.lock.namespace,
+                                                         leader_election_record)
+
+        if update_status is False:
+            logging.info("{} failed to acquire lease".format(leader_election_record.holder_identity))
+            return False
+
+        self.observed_record = leader_election_record
+        self.observed_time_milliseconds = int(time.time() * 1000)
+        logging.info("leader {} has successfully acquired lease".format(leader_election_record.holder_identity))
+        return True
--- a/leaderelection/leaderelection_test.py
+++ b/leaderelection/leaderelection_test.py
@@ -1,0 +1,270 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from . import leaderelection
+from .leaderelectionrecord import LeaderElectionRecord
+from kubernetes.client.rest import ApiException
+from . import electionconfig
+import unittest
+import threading
+import json
+import time
+import pytest
+
+thread_lock = threading.RLock()
+
+class LeaderElectionTest(unittest.TestCase):
+    def test_simple_leader_election(self):
+        election_history = []
+        leadership_history = []
+
+        def on_create():
+            election_history.append("create record")
+            leadership_history.append("get leadership")
+
+        def on_update():
+            election_history.append("update record")
+
+        def on_change():
+            election_history.append("change record")
+
+        mock_lock = MockResourceLock("mock", "mock_namespace", "mock", thread_lock, on_create, on_update, on_change, None)
+
+        def on_started_leading():
+            leadership_history.append("start leading")
+
+        def on_stopped_leading():
+            leadership_history.append("stop leading")
+
+        # Create config 4.5 4 3
+        config = electionconfig.Config(lock=mock_lock, lease_duration=2.5,
+                                       renew_deadline=2, retry_period=1.5, onstarted_leading=on_started_leading,
+                                       onstopped_leading=on_stopped_leading)
+
+        # Enter leader election
+        leaderelection.LeaderElection(config).run()
+
+        self.assert_history(election_history, ["create record", "update record", "update record", "update record"])
+        self.assert_history(leadership_history, ["get leadership", "start leading", "stop leading"])
+
+    def test_leader_election(self):
+        election_history = []
+        leadership_history = []
+
+        def on_create_A():
+            election_history.append("A creates record")
+            leadership_history.append("A gets leadership")
+
+        def on_update_A():
+            election_history.append("A updates record")
+
+        def on_change_A():
+            election_history.append("A gets leadership")
+
+        mock_lock_A = MockResourceLock("mock", "mock_namespace", "MockA", thread_lock, on_create_A, on_update_A, on_change_A, None)
+        mock_lock_A.renew_count_max = 3
+
+        def on_started_leading_A():
+            leadership_history.append("A starts leading")
+
+        def on_stopped_leading_A():
+            leadership_history.append("A stops leading")
+
+        config_A = electionconfig.Config(lock=mock_lock_A, lease_duration=2.5,
+                                         renew_deadline=2, retry_period=1.5, onstarted_leading=on_started_leading_A,
+                                         onstopped_leading=on_stopped_leading_A)
+
+        def on_create_B():
+            election_history.append("B creates record")
+            leadership_history.append("B gets leadership")
+
+        def on_update_B():
+            election_history.append("B updates record")
+
+        def on_change_B():
+            leadership_history.append("B gets leadership")
+
+        mock_lock_B = MockResourceLock("mock", "mock_namespace", "MockB", thread_lock, on_create_B, on_update_B, on_change_B, None)
+        mock_lock_B.renew_count_max = 4
+
+        def on_started_leading_B():
+            leadership_history.append("B starts leading")
+
+        def on_stopped_leading_B():
+            leadership_history.append("B stops leading")
+
+        config_B = electionconfig.Config(lock=mock_lock_B, lease_duration=2.5,
+                                         renew_deadline=2, retry_period=1.5, onstarted_leading=on_started_leading_B,
+                                         onstopped_leading=on_stopped_leading_B)
+
+        mock_lock_B.leader_record = mock_lock_A.leader_record
+
+        threading.daemon = True
+        # Enter leader election for A
+        threading.Thread(target=leaderelection.LeaderElection(config_A).run()).start()
+
+        # Enter leader election for B
+        threading.Thread(target=leaderelection.LeaderElection(config_B).run()).start()
+
+        time.sleep(5)
+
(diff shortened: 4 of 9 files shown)


Invictus17:
Hi @mbohlool @roycaihw , can you please review my PR?

yliaog:
thanks for the PR.

could you please keep the file naming style consistent with the existing code? e.g. LeaderElection is better to switched to leaderelection

Invictus17:
> thanks for the PR.
> 
> could you please keep the file naming style consistent with the existing code? e.g. LeaderElection is better to switched to leaderelection

Sure. I've pushed the changes.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,237 @@
...
+            print(self.leaderElectionRecord['holderIdentity'], "failed to lead")
+            # the user is notified
+            print(self.getLatestLeader())
+            # A new leader has been established
+            self.follow(scheduler)
+        else:
+            # Make sure thread that runs onStoppedLeading callback is stopped
+            if self.OnStoppedLeadingThread:
+                self.OnStoppedLeadingThread.stop()
+                self.OnStoppedLeadingThread.join()
```
roycaihw, line 118: `onStoppedLeading` is not used. Could you demonstrate how a user should react to `OnStoppedLeadingThread` in the example?

leaderelection/threadingwithexception.py
```diff
@@ -0,0 +1,31 @@
+import threading
+import ctypes
```
roycaihw, line 2: I'm not very excited if we have to add C compatibility to achieve this

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,237 @@
...
+            # kill the OnLeading callback
+            self.leaderFunctionThread.stop()
+            self.leaderFunctionThread.join()
+
+            # create OnStoppedLeading thread
+            self.OnStoppedLeadingThread = threadingwithexception.thread(self.electionConfig.onStoppedLeading)
+            self.OnStoppedLeadingThread.start()
+
+            # Start to follow
+            self.follow(scheduler)
```
roycaihw, line 156: In your implementation, a program can keep switching between being a leader and a follower. However reading the client-go implementation, when the leader fails to renew the lease, it [gives up and doesn't re-join as a follower](https://github.com/kubernetes/kubernetes/blob/f5334fad483856c41ceb1031036a96e96c32fcee/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L278-L279). Should we keep the client-go behavior and let the user decide what to do when a leader loses the lease?

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,237 @@
...
+    def leaderElectorRecord(self, electionConfig):
+        return {"holderIdentity": str(electionConfig.lock.identity),
+                "leaseDurationSeconds": str(electionConfig.leaseDuration),
+                "acquireTime": str(datetime.datetime.now())}
+
+    # Returns latest leader
+    def getLatestLeader(self):
+        getStatus, getResponse = self.electionConfig.lock.Get(name=self.electionConfig.lock.name, namespace=self.electionConfig.lock.namespace)
+        if getStatus:
+            return "leader is " + str(ast.literal_eval(getResponse.metadata.annotations[self.electionConfig.lock.LeaderElectionRecordAnnotationKey])['holderIdentity'])
```
roycaihw, line 56: why do we need `literal_eval`?

leaderelection/threadingwithexception.py
Invictus17: I can look into a different way to kill threads.

leaderelection/leaderelection.py
Invictus17: Sure. I can change the code to have a leader exit and close if failed to update lease.

leaderelection/leaderelection.py
roycaihw: nvm, I misread the thread logic

leaderelection/leaderelection.py
Invictus17: getResponse.metadata.annotations[self.electionConfig.lock.LeaderElectionRecordAnnotationKey] returns a string. 
literal_eval converts that string to a dictionary so that ['holderIdentity'] can be easily accessed.

Invictus17:
Hi @roycaihw , Thanks for your review. Based on your comments I've pushed the suggested changes.

changes:  
1) Coding style - Updated variable and function names.  
2) Threading with traces - This implementation does not depend on ctypes.
3) Leader exits - The leader does not become a follower if it fails to update lease. It now exits after running onStoppedLeading().

leaderelection/threadingwithtrace.py
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+
+import sys
+import trace
+import threading
+import time
+
+
+# citing source: https://www.geeksforgeeks.org/python-different-ways-to-kill-a-thread/
+class thread_with_trace(threading.Thread):
```
yliaog, line 22: it would be simpler to use https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor than working with threads directly.

it is added in python 3, but i guess it is ok since python 2 is deprecated anyway, you don't have to support python 2.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,257 @@
...
+            print(self.leader_election_record['holderIdentity'], "failed to lead")
+            # the user is notified
+            print(self.getlatest_leader())
+            # A new leader has been established
+            self.follow(scheduler)
+        else:
+            # Make sure thread that runs onstopped_leading callback is stopped
+            if self.onstopped_leadingthread:
+                self.onstopped_leadingthread.stop()
+                self.onstopped_leadingthread.join()
```
yliaog, line 133: i don't think you need to 'stop', then 'join', i.e., wait for it to stop. 

it's better to implement a simpler leaderelection:
1. right at the start, try to acquire lease to be the leader
2. if not yet a leader, periodically check if it can acquire the lease to be leader
3. if it becomes the leader, call the hook on started_leading
4. try to maintain the leadership by renewing the lease
5. if fail to renew the lease, call the hook on_stopped_leading
6. done with the leaderelection, return

NOTE: the lifecycle of one leaderelection run has two possibilities:
a) it always blocks, waiting to be a leader, but never succeeds
b) it becomes a leader, lead for sometime, then stop leading and return

It is simpler because during one leaderelection run, it can be a leader at most *once*. if become a leader, then somehow lose the leadership, then the whole leaderelection returns. it leave to the caller of the leaderelection to decide what to do after that, the caller may exit completely, or the caller may choose to run another leaderelection.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,257 @@
...
+        get_status, get_response = self.election_config.lock.get(name=self.election_config.lock.name, namespace=self.election_config.lock.namespace)
+        if get_status:
+            return "leader is " + str(ast.literal_eval(get_response.metadata.annotations[self.election_config.lock.leader_electionrecord_annotationkey])['holderIdentity'])
+        else:
+            return "could not get the latest leader"
+
+    # Point of entry to Leader election
+    def run(self):
+        # Try to create/ acquire a lock
+        self.try_acquire_or_renew()
```
yliaog, line 78: better to check out the java implementation, https://github.com/kubernetes-client/java/blob/master/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java

It uses only three threads, or more precisely, three threadpools, but each pool has only a single thread.

leaderelection/resourcelock/configmaplock.py
```diff
@@ -0,0 +1,74 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kubernetes.client.rest import ApiException
+from kubernetes import client, config
+from kubernetes.client.api_client import ApiClient
+
+
+class ConfigMapLock:
```
yliaog, line 20: maybe not now, but unittests need to be added before merging the PR

leaderelection/example.py
```diff
@@ -0,0 +1,46 @@
...
+
+import uuid
+import leaderelection
+from kubernetes import client, config
+from resourcelock.configmaplock import ConfigMapLock
+import electionconfig
+from datetime import timedelta
+
+# Authenticate using config file
+config.load_kube_config(config_file=r"D:\Kubernetes open source - Python client\Go example\kubeconfig.txt")
```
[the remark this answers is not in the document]
Invictus17: sorry, I forgot to leave that out. Updated now.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def try_acquire_or_renew(self):
+        # Check if lock is created
+        lock_status, lock_response = self.election_config.lock.get(self.election_config.lock.name, self.election_config.lock.namespace)
+
+        # create a default Election record for this candidate
+        self.leader_election_record = self.leaderelector_record(self.election_config)
```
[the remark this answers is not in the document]
Invictus17: Thank you for pointing it out. I'll update it in the following commit.

leaderelection/leaderelection.py
Invictus17: Thank you for pointing this out. I should have removed this piece of code when I updated the code for a leader to not follow after losing leadership and exit. Other than that, the leader election logic is identical to what you've described. I'll also make sure that the program let's the user decide whether to exit or run for election again. I'll handle these changes in the next commit.

leaderelection/threadingwithexception.py
cliffburdick: You should be able to use stop flags and let the client check for it.

leaderelection/leaderelection.py
cliffburdick: > `getResponse.metadata.annotations[self.electionConfig.lock.LeaderElectionRecordAnnotationKey]` returns a string.
> `literal_eval `converts that string to a dictionary so that `['holderIdentity']` can be easily accessed.

Would json.loads() work or is it not JSON?

leaderelection/threadingwithexception.py
Invictus17: > You should be able to use stop flags and let the client check for it.

Yes, that's similar to the latest update using traces(this commit is outdated now) but I also want to look into https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor as suggested by @yliaog in her review.

leaderelection/leaderelection.py
Invictus17: > > `getResponse.metadata.annotations[self.electionConfig.lock.LeaderElectionRecordAnnotationKey]` returns a string.
> > `literal_eval `converts that string to a dictionary so that `['holderIdentity']` can be easily accessed.
> 
> Would json.loads() work or is it not JSON?

No, it's not JSON.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,257 @@
...
+        self.oldelection_record = None
+
+        # islease_updated_follower is used by a follower candidate to mark if the leader has updated the lock object
+        self.islease_updated_follower = False
+        
+        # leaderfunction_thread contains the thread object for onstarted_leading
+        self.leaderfunction_thread = None
+        
+        # onstopped_leadingthread contains the thread object for onstopped_leading
+        self.onstopped_leadingthread = None
```
roycaihw, line 59: on_stopped_leading should not be run in a separate thread

Invictus17:
Hi @roycaihw @yliaog ,
Thank you for your review. I have updated my code based on your comments and discussions. The updates are:
1. Not killing threads and making the logic simpler, similar to the initial Java client.
2. OnStoppedLeading() is not run in a sub-thread.
3. Added a README.
4. Using logging and minor changes pointed out during review.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,243 @@
...
+            sys.exit("argument config not passed")
+
+        # The configuration set for this candidate
+        self.election_config = election_config
+        # The election record kept by this candidate
+        self.leader_election_record = None
+
+        # updatelease_schedulerId variable stores the scheduler object id for the update_lease schedule that is repeated
+        # every retry_period seconds by the leader
+        self.updatelease_schedulerId = None
```
roycaihw, line 36: is this unused?

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,243 @@
...
+
+                # Start leading and call OnStartedLeading()
+                self.leaderfunction_thread = threading.Thread(target=self.election_config.onstarted_leading, daemon=True)
+                self.leaderfunction_thread.start()
+
+                # Schedule the function of updating the lock
+                scheduler_leader = sched.scheduler(time.time, time.sleep)
+                self.lead(scheduler_leader)
+
+    def transition_follower_to_leader(self):
```
roycaihw, line 120: we don't need to distinguish `transition_follower_to_leader` and `lead`, same for `follow` and `try_acquire`. These complexity might be useful if we wanted to convert a leader back to a follower

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,243 @@
...
+            # compare timestamps
+            if self.oldelection_record['acquireTime'] != current_election_record['acquireTime']:
+                # Leader has updated lease continue following
+                self.islease_updated_follower = True
+
+            self.oldelection_record = current_election_record
+
+
+            # keep checking for lease updates every retry_period seconds
+            self.followerlease_checkscheduler = scheduler.enter(int(self.election_config.retry_period), 1, self.check_lease_updates, (scheduler,))
```
roycaihw, line 242: previously I thought the scheduler was like the `wait.Until` method that client-go uses, where the scheduling and the real logic are isolated. I found this implementation (scheduling the real logic recursively) hard to follow. What's the benefit against using a simple while loop + sleep like the java client?

leaderelection/leaderelection.py
Invictus17: No it is used:
https://github.com/Invictus17/python-base/blob/master/leaderelection/leaderelection.py#L138
https://github.com/Invictus17/python-base/blob/master/leaderelection/leaderelection.py#L182

leaderelection/leaderelection.py
Invictus17: I don't think either has a benefit over the other. Would you suggest using a while + sleep?

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,243 @@
...
+    def try_acquire_or_renew(self):
+        # Check if lock is created
+        lock_status, lock_response = self.election_config.lock.get(self.election_config.lock.name, self.election_config.lock.namespace)
+
+
+
+        # If a lock is already created with that name
+        if lock_status:
+            logging.info("{} is a follower".format(self.election_config.lock.identity))
+            scheduler_follower = sched.scheduler(time.time, time.sleep)
```
yliaog, line 87: there are a total of 4 of these: sched.scheduler(time.time, time.sleep) in the code. would one scheduler be sufficient? would it be better to create just one in the class __init__?

leaderelection/README.md
```diff
@@ -0,0 +1,11 @@
...
+This example demonstrates how to use the leader election library.
+
+## Running
+Run the following command in multiple separate terminals preferably an odd number. Each running process uses a unique identifier displayed when it starts to run.
+When a program runs, it can either be a leader or a follower. The user will be prompted about the status of the candidate and transitions. 
+
+### Command to run
+```python example.py```
+
+Now kill the existing leader. You will see from the terminal outputs that one of the remaining two processes will be elected as the new leader.
```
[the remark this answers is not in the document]
Invictus17: Thanks for pointing it out. I'll update it.

leaderelection/leaderelection.py
Invictus17: > there are a total of 4 of these: sched.scheduler(time.time, time.sleep) in the code. would one scheduler be sufficient? would it be better to create just one in the class **init**?

Yes, one would do, I'll update it. @roycaihw had a different opinion about using the sched module. Do you have similar thoughts about it?
https://github.com/kubernetes-client/python-base/pull/206#discussion_r471037636

yliaog: Yes, i agree with his comment. i think it's easier to understand to have the program structure like this:

1. try to acquire lease to be leader in main thread
2. if  get lease, then run lead in a separate thread
3. keep renewing lease in main thread
4. if failed renewing lease, then stop leading in main thread

Invictus17:
Hi @roycaihw @yliaog ,
Thank you for your review. I have updated my code based on your comments and discussions. The updates are:
1. Switched from using sched module to a while + sleep approach.
2. Made the code consistent with the other clients.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,156 @@
...
+    def try_acquire_or_renew(self):
+        # Check if lock is created
+        lock_status, lock_response = self.election_config.lock.get(self.election_config.lock.name, self.election_config.lock.namespace)
+
+        # create a default Election record for this candidate
+        leader_election_record = self.leaderelector_record(self.election_config)
+        
+        # If a lock is already created with that name
+        if lock_status:
+            old_election_record = ast.literal_eval(lock_response.metadata.annotations[self.election_config.lock.leader_electionrecord_annotationkey])
```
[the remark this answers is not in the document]
yliaog: also ast.literal_eval could throw an exception, better to catch it

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,156 @@
...
+                logging.info("Leader has switched to {}".format(old_election_record['holderIdentity']))
+
+
+            if old_election_record != self.observed_record:
+                self.observed_record = old_election_record
+                self.observed_time_milliseconds = int(time.time()*1000)
+
+
+            # If This candidate is not the leader and lease duration is yet to finish
+            if str(self.election_config.lock.identity) != self.observed_record['holderIdentity'] and self.observed_time_milliseconds + self.election_config.lease_duration*1000 > int(time.time()*1000):
```
yliaog, line 121: self.observed_time_milliseconds is from this candidate, but it should be from the election record, no?

also self.observed_time_milliseconds is updated to the current time at line 117, so it will always be the current time, and the lease duration will never expire

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,156 @@
...
+                logging.info("yet to finish lease_duration, lease held by {} and has not expired".format(old_election_record['holderIdentity']))
+                return False
+
+            # If this candidate is the Leader
+            if str(self.election_config.lock.identity) == self.observed_record['holderIdentity']:
+                # Leader sets acquireTime
+                leader_election_record['acquireTime'] = self.observed_record['acquireTime']
```
yliaog, line 128: self.observed_record['acquireTime'] is not updated, it is still the old_election_record's acquireTime

leaderelection/leaderelection.py
Invictus17: > self.observed_time_milliseconds is from this candidate, but it should be from the election record, no?
No, the java and go clients are also keeping a local record of 'observed_time_milliseconds' and not referring to the election record
https://github.com/kubernetes-client/java/blob/master/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java#L269-L315

 
> also self.observed_time_milliseconds is updated to the current time at line 117, so it will always be the current time, and the lease duration will never expire
It is only updated if a follower identifies that the lock object has been updated by a leader.
       if old_election_record != self.observed_record:
             # Update self.observed_time_milliseconds & self.observed_record

In case a leader fails to update the lock, self.observed_time_milliseconds will not be updated to current time & after a period of 'leaseDuration' a follower will try to update the lock. 
https://github.com/kubernetes-client/java/blob/master/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java#L267-L279

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,156 @@
...
+            old_election_record = ast.literal_eval(lock_response.metadata.annotations[self.election_config.lock.leader_electionrecord_annotationkey])
+
+            # Report transitions
+            if self.observed_record and self.observed_record['holderIdentity'] != old_election_record['holderIdentity']:
+                logging.info("Leader has switched to {}".format(old_election_record['holderIdentity']))
+
+
+            if old_election_record != self.observed_record:
+                self.observed_record = old_election_record
+                self.observed_time_milliseconds = int(time.time()*1000)
```
roycaihw, line 117: capture the current time (e.g. `now = time.time()`) at the beginning of this function, and keep using `now`, instead of calling time.time() multiple times.

leaderelection/leaderelection.py
roycaihw: this line is correct I think. It prevents leader_election_record['acquireTime'] to be "now"

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,156 @@
...
+                return False
+
+            # If this candidate is the Leader
+            if str(self.election_config.lock.identity) == self.observed_record['holderIdentity']:
+                # Leader sets acquireTime
+                leader_election_record['acquireTime'] = self.observed_record['acquireTime']
+
+            # Update object with latest election record
+            lock_response.metadata.annotations[self.election_config.lock.leader_electionrecord_annotationkey] = str(leader_election_record)
+            update_status, update_response = self.election_config.lock.update(self.election_config.lock.name, self.election_config.lock.namespace, lock_response)
```
roycaihw, line 132: this is not generalized for different types of locks. Better to make leader_election_record a class or a dict, and pass it directly to the lock methods

roycaihw: the leaderelection logic doesn't need to know about `LeaderElectionRecordAnnotationKey`. It's a implementation detail for the lock

leaderelection/leaderelection.py
yliaog: ok, the "> int(time.time()*1000):" in the above line is incorrect. time.time() would return the current time, what should be used instead is the time at the start of the function try_acquire_or_renew (https://github.com/kubernetes-client/java/blob/6a2a60ad2ad75fb127874797ea910b90e4a80651/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java#L236)

leaderelection/leaderelection.py
yliaog: ok. i was thinking about setting the 'acquireTime' at the time when the lease is acquired, currently, the 'acquireTime' is set at the line: 104: leader_election_record = self.leaderelector_record(self.election_config)

it probably does not matter though, as there is no wait or loop inside the function try_acquire_or_renew

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,182 @@
...
+            return False
+
+        # If this candidate is the Leader
+        if str(self.election_config.lock.identity) == self.observed_record['holderIdentity']:
+            # Leader sets acquireTime
+            leader_election_record['acquireTime'] = self.observed_record['acquireTime']
+
+        return self.update_lock(lock_response, leader_election_record)
+
+    def update_lock(self, lock_response, leader_election_record):
```
roycaihw, line 168: `lock_response` can be a field stored in the lock, since it was returned from `lock.get()` and never changed before we passed it back to `lock.update()`. See [client-go](https://github.com/kubernetes/client-go/blob/57f65489c11a4d4a978c93ab369c38f53156888a/tools/leaderelection/resourcelock/configmaplock.go#L41) as an example.

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,182 @@
...
+        # check for any key, value errors in the record
+        try:
+            old_election_record = ast.literal_eval(lock_record)
+            if (old_election_record['holderIdentity'] == '' or old_election_record['leaseDurationSeconds'] == ''
+                    or old_election_record['acquireTime'] == '' or old_election_record['renewTime'] == ''):
+                # try to update lock with proper annotation and election record
+                return self.update_lock(lock_response, leader_election_record)
+        except:
+            # try to update lock with proper annotation and election record
+            return self.update_lock(lock_response, leader_election_record)
```
roycaihw, line 144: I think it's better to assume `lock_record` is a class object, instead of a string here. Each lock type can have its own unmarshalling implementation (e.g. ConfigMap and Endpoints unmarshal their annotations, while Lease unmarshalls its spec). 

This will also allow the short-circuit logic here to be as simple as [client-java](https://github.com/kubernetes-client/java/blob/abea7f4dbf9bf523a4e94349129309da09e317e0/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java#L259-L264).

leaderelection/resourcelock/configmaplock.py
```diff
@@ -0,0 +1,129 @@
...
+    def create(self, name, namespace, election_record):
+        """
+        :param electionRecord: Annotation string
+        :param name: Name of the configmap object to be created
+        :param namespace: Namespace in which the configmap object is to be created
+        :return: 'True' if object is created else 'False' if failed
+        """
+        body = client.V1ConfigMap(
+            metadata={"name": name,
+                      "annotations": {self.leader_electionrecord_annotationkey: json.dumps(self.get_lock_dict(election_record))}})
```
roycaihw, line 83: `self.get_lock_dict` always updates `self.lock_record`. I think we don't want to update the lock record if the create fails. Same for `update()`.

roycaihw (commented):
Thanks for generalizing the leader_election_record. Overall the implementation looks good to me. Please add a test.

leaderelection/resourcelock/configmaplock.py
roycaihw: nvm, was thinking of `self.configmap_reference`

leaderelection/leaderelection.py
```diff
@@ -101,82 +93,76 @@ def try_acquire_or_renew(self):
...
         # create a default Election record for this candidate
-        leader_election_record = self.leaderelector_record(self.election_config, now)
+        leader_election_record = LeaderElectionRecord(self.election_config.lock.identity,
+                                                     str(self.election_config.lease_duration), str(now), str(now))
 
         # A lock is not created with that name, try to create one
         if not lock_status:
-            if ast.literal_eval(lock_response.body)['code'] != 404:
-                logging.info("Error retrieving resource lock {} as {}".format(self.election_config.lock.name, lock_response.reason))
+            if json.loads(old_election_record.body)['code'] != 404:
```
yliaog, line 105: please avoid magic constant 404, instead use NOT_FOUND

leaderelection/leaderelection.py
```diff
@@ -0,0 +1,168 @@
...
+        self.observed_time_milliseconds = 0
+
+    # Point of entry to Leader election
+    def run(self):
+        # Try to create/ acquire a lock
+        if self.acquire():
+            logging.info("{} successfully acquired lease".format(self.election_config.lock.identity))
+
+            # Start leading and call OnStartedLeading()
+            threading.Thread(target=self.election_config.onstarted_leading, daemon=True).start()
```
yliaog, line 47: please add some documentation to the class to describe the threading behavior. i.e. the thread is running until onstarted_leading returns if it does return, it's not safe to run this leader election more than once in a process, etc.

Invictus17:
@yliaog @roycaihw In my latest push I've added the tests but the travis ci report is indicating a few failures. Any idea why some of these tests are failing?
test result: https://travis-ci.org/github/kubernetes-client/python-base/builds/751412124

yliaog:
https://github.com/kubernetes-client/python-base/pull/222 fixed the configmap test failure

leaderelection/leaderelection_test.py
```diff
@@ -0,0 +1,247 @@
...
+
+        def on_stopped_leading_A():
+            leadership_history.append("A stops leading")
+
+        config_A = electionconfig.Config(lock=mock_lock_A, lease_duration=2.5,
+                                         renew_deadline=2, retry_period=1.5, onstarted_leading=on_started_leading_A,
+                                         onstopped_leading=on_stopped_leading_A)
+
+        # Enter leader election
+        leaderelection.LeaderElection(config_A).run()
```
roycaihw, line 87: Since `run()` blocks until the leader election ends, B won't start until A hits `renew_count_max`. Could you make the two clients run in parallel to make sure the leader election handles concurrency properly?

leaderelection/leaderelection_test.py
```diff
@@ -0,0 +1,247 @@
...
+        leaderelection.LeaderElection(config).run()
+
+        self.assert_history(election_history,
+                            ["create record",
+                             "try update record",
+                             "update record",
+                             "try update record",
+                             "update record",
+                             "try update record",
+                             "try update record"])
```
roycaihw, line 172: I had some trouble understanding what this test (`test_Leader_election_with_renew_deadline `) does. Could you add some comments to explain how it's related to `renew_deadline` and what's the expected behavior? Thanks

Invictus17: Thanks @roycaihw, sure I'll add some comments. The expected behavior is to check if the leader stops leading if it fails to update the lock after the renew_deadline.

leaderelection/leaderelection_test.py
```diff
@@ -0,0 +1,267 @@
...
+    """Expected behavior: to check if the leader stops leading if it fails to update the lock within the renew_deadline
+    and stops leading after finally timing out. The difference between each try comes out to be approximately the sleep
+    time.
+    Example:
+    create record:  0s
+    on try update:  1.5s
+    on update:  zzz s
+    on try update:  3s
+    on update: zzz s 
+    on try update:  4.5s
```
roycaihw, line 147: Before this try, the timeout was set to be 4.5 + renew_deadline = 6.5s. After two failed tries (w/ sleep), the leader exited at 4.5+1.5*2=7.5s.

roycaihw:
LGTM. Please squash the commits into reasonable parts.

leaderelection/leaderelection_test.py
```diff
@@ -0,0 +1,267 @@
...
+
+class MockResourceLock:
+    def __init__(self, name, namespace, identity, on_create=None, on_update=None, on_change=None, on_try_update=None):
+        self.leader_record = []
+        self.renew_count = 0
+        self.renew_count_max = 4
+        self.name = name
+        self.namespace = namespace
+        self.identity = str(identity)
+        self.lock = threading.RLock()
```
roycaihw, line 214: is this lock per MockResourceLock? I'd expect it to be shared between all MockResourceLocks in one test

Invictus17: No, it should be shared. I'll update it. Thanks!

roycaihw:
/lgtm
/approve

Thanks!

PR #208, opened 2020-08-19
Support file-ish objects in config loading

### LINKED ISSUE #207: Config isn't able to take a string, or StringIO object
Currently I use Rancher and K8s together. I can currently get my kubeconfig dynamically from a Rancher endpoint. 

However, the kubernetes-client doesn't support any capability to use a kubeconfig from string, or even StringIO because a 'split' on the path is required.

The python client should support a kubeconfig via string for situations in which a filesystem either isn't available, or isn't viable.

ryphon:
Closes #207

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -669,21 +669,15 @@ class KubeConfigMerger:
         self.config_files = {}
         self.config_merged = None
         if hasattr(paths, 'read'):
-            self.load_config_from_fileish(paths)
+            self._load_config_from_file_like_object(paths)
         else:
-            for path in paths.split(ENV_KUBECONFIG_PATH_SEPARATOR):
-                if path:
-                    path = os.path.expanduser(path)
-                    if os.path.exists(path):
-                        self.paths.append(path)
-                        self.load_config(path)
-            self.config_saved = copy.deepcopy(self.config_files)
+            self._load_config_from_file_path(paths)
 
     @property
     def config(self):
         return self.config_merged
 
-    def load_config_from_fileish(self, string):
+    def _load_config_from_file_like_object(self, string):
         if hasattr(string, 'getvalue'):
             config = yaml.safe_load(string.getvalue())
         else:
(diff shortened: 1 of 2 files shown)


ryphon:
[working.txt](https://github.com/kubernetes-client/python-base/files/5098958/working.txt)

File attached of a python console loading a config from local file into a string, and then that string being used as config for the api client.

config/kube_config.py
```diff
@@ -667,19 +668,31 @@ def __init__(self, paths):
...
+        if hasattr(paths, 'read'):
+            self.load_config_from_fileish(paths)
+        else:
+            for path in paths.split(ENV_KUBECONFIG_PATH_SEPARATOR):
+                if path:
+                    path = os.path.expanduser(path)
+                    if os.path.exists(path):
+                        self.paths.append(path)
+                        self.load_config(path)
+            self.config_saved = copy.deepcopy(self.config_files)
```
palnabarun, line 680: How about pushing the code path for paths to a separate internal function (say `_load_config_from_file_path`) as well? I think readability will increase that way.

config/kube_config.py
```diff
@@ -667,19 +668,31 @@ def __init__(self, paths):
...
 
     @property
     def config(self):
         return self.config_merged
 
+    def load_config_from_fileish(self, string):
```
palnabarun, line 686: How does renaming the method to `_load_config_from_file_like_object` sound? (`_` since this is not supposed to be used externally)

palnabarun (changes requested):

config/kube_config.py
ryphon: Sounds like a great point, will get that in shortly.

config/kube_config.py
ryphon: I like this as well, will get this in shortly as well!

roycaihw:
There is a test failure https://travis-ci.org/github/kubernetes-client/python-base/builds/722838842:

```
_______ TestKubeConfigLoader.test_load_kube_config_from_file_like_object _______
self = <config.kube_config_test.TestKubeConfigLoader testMethod=test_load_kube_config_from_file_like_object>
    def test_load_kube_config_from_file_like_object(self):
        expected = FakeConfig(host=TEST_HOST,
                              token=BEARER_TOKEN_FORMAT % TEST_DATA_BASE64)
        config_file_like_object = io.StringIO()
>       config_file_like_object.write(yaml.safe_dump(self.TEST_KUBE_CONFIG))
E       TypeError: unicode argument expected, got 'str'
kubernetes/base/config/kube_config_test.py:1265: TypeError
```

Please take a look. LGTM otherwise.

/approve

ryphon:
Ah silly me, I didn't realize I also had to write for 2.7. I'll get another commit in the next couple days to address.

ryphon:
Not sure why travis isn't reporting, but the tests now pass. @roycaihw @palnabarun @yliaog

Issue #192, opened 2020-04-30
ws_client.py could not handle binary as stdout


walthhy:
k8s python client: 11.0.0
python 3

We're using following block copying files/dirs from pod to local, however, ws_client would decode(utf-8) data frame from socket into str, which would break the function. Enhancement needed for function ```def update(self, timeout=0)``` to store data based on type into channels. 

```python
    exec_command = ['tar', 'cf', '-', '/var']

    with TemporaryFile() as tar_buffer:
        resp = stream(api_instance.connect_get_namespaced_pod_exec, pod_name, namespace,
                      command=exec_command,
                      stderr=True, stdin=False,
                      stdout=True, tty=False,
                      _preload_content=False)

        while resp.is_open():
            resp.update(timeout=1)
            if resp.peek_stdout():
                out=resp.read_stdout()
                tar_buffer.write(out)
            if resp.peek_stderr():
                print("STDERR: %s" % resp.read_stderr())

        resp.close()

        tar_buffer.flush()
        tar_buffer.seek(0)

        with tarfile.open(fileobj=tar_buffer, mode='r:') as tar:
            print('members', tar.getmembers())

```

roycaihw:
We should only decode the response if [`opcode == ABNF.OPCODE_TEXT`](https://github.com/websocket-client/websocket-client/blob/29c15714ac9f5272e1adefc9c99b83420b409f63/websocket/_core.py#L315-L316) (websocket-client module). Currently we are decoding binary payload as well: https://github.com/kubernetes-client/python-base/blob/bf5c59987122fd35e9bb14a939f3db94e55b30bf/stream/ws_client.py#L207-L210

We also need a e2e test case for this. I will work on this when I get a chance, or someone else may pick it up. 

/help

iciclespider:
Here is a total curve ball observation.

I think the wrapping of the pod exec api needs to be completely reimplemented following the python subprocess.Popen constructor calling sequence as close as possible. This new implementation should return an object that then acts as close as possible to an instance of subprocess.Popen. This is a well thought out library for interacting with processes. Then python programmers that know how to interact with subprocess.Popen will be up and running doing the same with pod exec.

Then in response to this specific issue, subprocess.Popen by default handles stdin, stdout, and stderr as binary streams. The `text=True` argument specifies that these streams should be text streams.

PR #209, opened 2020-08-20
Fix decode in ws_client

pigletfly:
fix https://github.com/kubernetes-client/python-base/issues/192

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -12,21 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from kubernetes.client.rest import ApiException
-
-import certifi
 import collections
 import select
 import ssl
 import time
 
+import certifi
 import six
 import yaml
 
 from six.moves.urllib.parse import urlencode, urlparse, urlunparse
 from six import StringIO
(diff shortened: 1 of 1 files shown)


palnabarun:
@pigletfly Can you please rebase against `master` to resolve the conflicts? :slightly_smiling_face:

pigletfly:
@palnabarun rebased

stream/ws_client.py
```diff
@@ -173,7 +173,8 @@ def update(self, timeout=0):
             elif op_code == ABNF.OPCODE_BINARY or op_code == ABNF.OPCODE_TEXT:
                 data = frame.data
                 if six.PY3:
-                    data = data.decode("utf-8", "replace")
+                    if op_code == ABNF.OPCODE_TEXT:
+                        data = data.decode("utf-8", "replace")
```
iciclespider, line 177: FYI, `ABNF.OPCODE_TEXT` will never be sent from kubernetes. All channels either send or receive `ABNF.OPCODE_BINARY` on the server side.

roycaihw: I think we need a test to demonstrate if this fix works.

PR #210, opened 2020-08-24
Implement port forwarding.

iciclespider:
Looking for feed back if this looks on track as a way of implementing port forwarding.

See the example and unittests added in this PR https://github.com/kubernetes-client/python/pull/1237 for usage.

https://github.com/kubernetes-client/python/issues/166

### DIFF
--- a/stream/stream.py
+++ b/stream/stream.py
@@ -17,9 +17,12 @@ import functools
 from . import ws_client
 
 
-def _websocket_reqeust(websocket_request, api_method, *args, **kwargs):
+def _websocket_reqeust(websocket_request, force_kwargs, api_method, *args, **kwargs):
     """Override the ApiClient.request method with an alternative websocket based
     method and call the supplied Kubernetes API method with that in place."""
+    if force_kwargs:
+        for kwarg, value in force_kwargs.items():
+            kwargs[kwarg] = value
     api_client = api_method.__self__.api_client
     # old generated code's api client has config. new ones has configuration
     try:
@@ -34,4 +37,5 @@ def _websocket_reqeust(websocket_request, api_method, *args, **kwargs):
         api_client.request = prev_request
 
 
-stream = functools.partial(_websocket_reqeust, ws_client.websocket_call)
+stream = functools.partial(_websocket_reqeust, ws_client.websocket_call, None)
+portforward = functools.partial(_websocket_reqeust, ws_client.portforward_call, {'_preload_content':False})
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -322,19 +322,21 @@ class PortForward:
                 rlist.append(self.websocket)
                 if kubernetes_data:
                     wlist.append(self.websocket)
-            all_closed = True
+            local_all_closed = True
             for port in self.local_ports.values():
                 if port.python.fileno() != -1:
-                    if port.data:
-                        wlist.append(port.python)
-                        all_closed = False
+                    if self.websocket.connected:
+                        rlist.append(port.python)
+                        if port.data:
+                            wlist.append(port.python)
+                        local_all_closed = False
                     else:
-                        if self.websocket.connected:
-                            rlist.append(port.python)
-                            all_closed = False
+                        if port.data:
+                            wlist.append(port.python)
+                            local_all_closed = False
                         else:
                             port.python.close()
-            if all_closed and (not self.websocket.connected or not kubernetes_data):
+            if local_all_closed and not (self.websocket.connected and kubernetes_data):
                 self.websocket.close()
                 return
             r, w, _ = select.select(rlist, wlist, [])
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -249,6 +249,10 @@ class PortForward:
             daemon=True
         ).start()
 
+    @property
+    def connected(self):
+        return self.websocket.connected
+
     def socket(self, port_number):
         if port_number not in self.local_ports:
             raise ValueError("Invalid port number")
@@ -276,8 +280,8 @@ class PortForward:
             s, self.python = socket.socketpair(socket.AF_UNIX, socket.SOCK_STREAM)
             # The self.socket half of the pair is used by the python application to send
             # and receive data to the eventual pod port. It is wrapped in the _Socket class
-            # because a socket pair is an AF_UNIX socket, not a AF_NET socket. This allows
-            # intercepting setting AF_INET socket options that would error against an AD_UNIX
+            # because a socket pair is an AF_UNIX socket, not a AF_INET socket. This allows
+            # intercepting setting AF_INET socket options that would error against an AF_UNIX
             # socket.
             self.socket = self._Socket(s)
             # Data accumulated from the websocket to be sent to the python application.
@@ -325,17 +329,17 @@ class PortForward:
             local_all_closed = True
             for port in self.local_ports.values():
                 if port.python.fileno() != -1:
-                    if self.websocket.connected:
-                        rlist.append(port.python)
-                        if port.data:
-                            wlist.append(port.python)
-                        local_all_closed = False
-                    else:
+                    if port.error or not self.websocket.connected:
                         if port.data:
                             wlist.append(port.python)
                             local_all_closed = False
                         else:
                             port.python.close()
+                    else:
+                        rlist.append(port.python)
+                        if port.data:
+                            wlist.append(port.python)
+                        local_all_closed = False
             if local_all_closed and not (self.websocket.connected and kubernetes_data):
                 self.websocket.close()
                 return
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -238,33 +238,51 @@ class PortForward:
 
         self.websocket = websocket
         self.local_ports = {}
-        for ix, local_remote in enumerate(ports):
-            self.local_ports[local_remote[0]] = self._Port(ix, local_remote[1])
+        for ix, port_number in enumerate(ports):
+            self.local_ports[port_number] = self._Port(ix, port_number)
+        # There is a thread run per PortForward instance which performs the translation between the
+        # raw socket data sent by the python application and the websocket protocol. This thread
+        # terminates after either side has closed all ports, and after flushing all pending data.
         threading.Thread(
-            name="Kubernetes port forward proxy", target=self._proxy, daemon=True
+            name="Kubernetes port forward proxy: %s" % ', '.join([str(port) for port in ports]),
+            target=self._proxy,
+            daemon=True
         ).start()
 
-    def socket(self, local_number):
-        if local_number not in self.local_ports:
+    def socket(self, port_number):
+        if port_number not in self.local_ports:
             raise ValueError("Invalid port number")
-        return self.local_ports[local_number].socket
+        return self.local_ports[port_number].socket
 
-    def error(self, local_number):
-        if local_number not in self.local_ports:
+    def error(self, port_number):
+        if port_number not in self.local_ports:
             raise ValueError("Invalid port number")
-        return self.local_ports[local_number].error
+        return self.local_ports[port_number].error
 
     def close(self):
         for port in self.local_ports.values():
             port.socket.close()
 
     class _Port:
-        def __init__(self, ix, remote_number):
-            self.remote_number = remote_number
+        def __init__(self, ix, port_number):
+            # The remote port number
+            self.port_number = port_number
+            # The websocket channel byte number for this port
             self.channel = bytes([ix * 2])
+            # A socket pair is created to provide a means of translating the data flow
+            # between the python application and the kubernetes websocket. The self.python
+            # half of the socket pair is used by the _proxy method to receive and send data
+            # to the running python application.
             s, self.python = socket.socketpair(socket.AF_UNIX, socket.SOCK_STREAM)
+            # The self.socket half of the pair is used by the python application to send
+            # and receive data to the eventual pod port. It is wrapped in the _Socket class
+            # because a socket pair is an AF_UNIX socket, not a AF_NET socket. This allows
+            # intercepting setting AF_INET socket options that would error against an AD_UNIX
+            # socket.
             self.socket = self._Socket(s)
+            # Data accumulated from the websocket to be sent to the python application.
             self.data = b''
+            # All data sent from kubernetes on the port error channel.
             self.error = None
 
         class _Socket:
@@ -285,8 +303,7 @@ class PortForward:
     def _proxy(self):
         channel_ports = []
         channel_initialized = []
-        python_ports = {}
-        rlist = []
+        local_ports = {}
         for port in self.local_ports.values():
             # Setup the data channel for this port number
             channel_ports.append(port)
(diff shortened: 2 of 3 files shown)


yliaog:
thanks for the PR. it looks good overall. 

could you please split this PR into two PRs, the first PR is for refactoring the existing code to make it reusable for port forwarding, the second PR is for port forwarding.

this makes it easier to review, and also easier to rollback or fix the portforward PR in case it has any issue.

iciclespider:
@yliaog  PR to refactor existing implementation in preparation for implementing port forwarding: #211

iciclespider:
@yliaog I rebased this PR against the merged master with the refactored code. I also updated the PR in https://github.com/kubernetes-client/python/pull/1237 with a `pod_portforward.py` example.

stream/ws_client.py
```diff
@@ -302,3 +441,34 @@ def websocket_call(configuration, _method, url, **kwargs):
...
+            for port in value.split(','):
+                try:
+                    port = int(port)
+                    if not (0 < port < 65536):
+                        raise ValueError
+                    ports.append(port)
+                except ValueError:
+                    raise ApiValueError("Invalid port number `" + str(port) + "`")
+    if not ports:
+        raise ApiValueError("Missing required parameter `ports`")
```
palnabarun, line 465: I think this logic can be abstracted out to another function with a simpler implementation like below.

```
def _valid_ports(ports):
    _ports = []

    for p in ports:
        try:
            p = int(p)
        except ValueError:
            raise ApiValueError("Invalid port number `{}`. Port number must be an integer".format(p))

        if (p < 0) or (p > 65536):
            raise ApiValueError("Invalid port number `{}`. Port number must be between 0 and 65536".format(p))

        _ports.append(p)

    return _ports

ports = query_params.get("ports", "")
if not ports:
    raise ApiValueError("Missing required parameter `ports`")

ports = ports.split(",")
ports = _valid_ports(ports)

# There is no check required here, since any unwanted port value would be checked in _valid_ports itself.
# We can be assured that we get a non-empty port list here.
```

palnabarun (changes requested):

stream/ws_client.py
iciclespider: This line will not work:
```
ports = query_params.get("ports", "")
```
query_params is not a dictionary, it an a list of two element tuples.

stream/ws_client.py
```diff
@@ -302,3 +441,34 @@ def websocket_call(configuration, _method, url, **kwargs):
...
+    """An internal function to be called in api-client when a websocket
+    connection is required for port forwarding. args and kwargs are the
+    parameters of apiClient.request method."""
+
+    query_params = kwargs.get("query_params")
+
+    ports = []
+    for key, value in query_params:
+        if key == 'ports':
+            for port in value.split(','):
```
yliaog, line 456: ports should be list of local_port:remote_port pairs

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+                # and is set when creating an SSLSocket from this socket.
+                if level == socket.IPPROTO_TCP and optname == socket.TCP_NODELAY:
+                    return
+                self._socket.setsockopt(level, optname, value)
+
+    # Proxy all socket data between the python code and the kubernetes websocket.
+    def _proxy(self):
+        channel_ports = []
+        channel_initialized = []
+        python_ports = {}
```
yliaog, line 288: what about call it local_port?

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+
+    def close(self):
+        for port in self.ports.values():
+            port.socket.close()
+
+    class _Port:
+        def __init__(self, ix, number):
+            self.number = number
+            self.channel = bytes([ix * 2])
+            s, self.python = socket.socketpair(socket.AF_UNIX, socket.SOCK_STREAM)
```
yliaog, line 265: what about renaming self.python to self.local?

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+            self.socket = self._Socket(s)
```
yliaog, line 266: what about renaming self.socket to self.remote_xxx?

yliaog: actually, why is sockepair needed? only one socket for each local port is needed, no?

stream/ws_client.py
yliaog: https://github.com/kubernetes/client-go/blob/master/tools/portforward/portforward.go

is the implementation for client-go

stream/stream.py
```diff
@@ -17,9 +17,12 @@
 from . import ws_client
 
 
-def _websocket_reqeust(websocket_request, api_method, *args, **kwargs):
+def _websocket_reqeust(websocket_request, force_kwargs, api_method, *args, **kwargs):
     """Override the ApiClient.request method with an alternative websocket based
     method and call the supplied Kubernetes API method with that in place."""
+    if force_kwargs:
+        for kwarg, value in force_kwargs.items():
+            kwargs[kwarg] = value
```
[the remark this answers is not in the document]
iciclespider: It must be supplied to the `connect_get_namespaced_pod_portforward` when it is called in the method so that the kubernetes client implementation between calling `connect_get_namespaced_pod_portforward` and calling `ws_client.portforward_call` returns the raw response, and does not try to return the json'ed version of what it expects to be a typical json reply.

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+            self.error = None
+
+        class _Socket:
+            def __init__(self, socket):
+                self._socket = socket
+
+            def __getattr__(self, name):
+                return getattr(self._socket, name)
+
+            def setsockopt(self, level, optname, value):
```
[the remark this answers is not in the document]
iciclespider: `http.client` makes this call right here: https://github.com/python/cpython/blob/8ca8a2e8fb068863c1138f07e3098478ef8be12e/Lib/http/client.py#L938

The code in this PR encounters this situation when socket.client_connection is monkey patched to return this connected socket in response to the `<pod-name>.<namespace>.kubernetes` dns names. This then allows the python code to use `http.client` or use `urllib` to make http requests that get routed to kubernetes pods.

It turns out that a socket returned from socketpair does not support the option being checked for below and throws an exception:
```
                if level == socket.IPPROTO_TCP and optname == socket.TCP_NODELAY:
```
Hence, the entire purpose of this `class _Socket` is to address this very issue. Otherwise the socket returned from socketpair could have been directly returned.

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+    def _proxy(self):
+        channel_ports = []
+        channel_initialized = []
+        python_ports = {}
+        rlist = []
+        for port in self.ports.values():
+            channel_ports.append(port)
+            channel_initialized.append(False)
+            channel_ports.append(port)
+            channel_initialized.append(False)
```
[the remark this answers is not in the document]
iciclespider: This is intentional to setup the data channel and the error channel. I will add a comment to make this clear.

stream/ws_client.py
iciclespider: Good catch. I'll adjust this parsing logic.

stream/ws_client.py
iciclespider: There are the following `socket`s involved:

1. The websocket's socket used to receive and send data to kubernetes. This is the `PortForward.websocket.sock` field. There is only one such socket involved.
2. The socket given to the running python code that the python code uses to send and receive data. This is currently what the `_Port.socket` field is. I called it `socket` because it is the socket supplied to the python code.
3. The socket used by the `PortForward._proxy` method to receive and send data to the python code that is using the `_Port.socket` socket. This is currently what the `_Port.python` code is. I called it `python` because it is the socket that `_proxy` is using to interact with the python code with.

The flow of data being sent from the python to kubernetes using this is:

* The python send data using the `_Port.socket` it has been supplied with.
* The `PortForward._proxy` method receives the above data using the `_Port.python` socket.
* The `PortForward._proxy` method sends the received data to `PortForward.websocket.sock` using the websocket channel that that port is associated with and using the websocket library to format the data.

The flow of data being sent from kubernetes to python using this is:

* The `PortForward._proxy` method receives data from kubernetes on the `PortForward.websocket.sock` socket and using the websocket library
* The `PortForward._proxy` method` sends the received data to the `_Port.python` socket associated with the websocket channel the data arrived on.
* The python code receives the above data on it's `_Port.socket` socket.

Hopefully the above makes clear what the socketpair is accomplishing. It is used to enable translating between the "raw" socket protocol the python is using, and the websocket based protocol that kubernetes is using, all without the python knowing what is going on.

So this still leaves what to call each end of the socketpair. I chose `_Port.socket` for the name of the socket used by the running python application to use as it's "socket". And I chose `_Port.python` for the name of the socket used by `PortForward._proxy` because it is using that socket to interact with the running python application.

I am will call them any name you prefer.

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+        threading.Thread(
+            name="Kubernetes port forward proxy", target=self._proxy, daemon=True
+        ).start()
+
+    def socket(self, port_number):
+        if port_number not in self.ports:
+            raise ValueError("Invalid port number")
+        return self.ports[port_number].socket
+
+    def error(self, port_number):
```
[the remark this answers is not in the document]
iciclespider: I just pushed the change to `error_channel`.

However, a little clarification. This is not really a "channel" that can send and receive data. From the client perspective, it is read only and cannot be written to. And the only time it is written to by kubernetes is when there is an error/exception case on the server.

See: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cri/streaming/portforward/websocket.go#L184-L197

stream/ws_client.py
yliaog: thanks for the clarification. here is my understanding, take a simple example, portforward from localhost:8080 to remote pod nginx port 80.

* there is just one websocket PortForward.websocket.sock used to send to/recv from pod nginx:80

* there is one local socket that binds to localhost:8080

1. listen on the socket (localhost:8080), wait for connection request
2. when there is a connection request, accept it, open a new connection, use that connection to  communicate with remote pod nginx:80 via the websocket

there could be multiple connection requests, hence multiple connections could be opened to handle them.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/portforward/portforward.go is the client-go's implementation.

i don't quite see how socketpair fits into the above flow.

iciclespider: Please look at the code example in this PR https://github.com/kubernetes-client/python/pull/1237

This implementation does not expose a port for other applications to connect to. Instead, it allows a python application to interact with the pod ports. There is not attempt to expose a port that other non-python executables can access in the same way kubectl exposes a port for other applications to use.

stream/ws_client.py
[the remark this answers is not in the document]
yliaog: ok

stream/ws_client.py
```diff
@@ -302,3 +441,56 @@ def websocket_call(configuration, _method, url, **kwargs):
...
+
+
+def portforward_call(configuration, _method, url, **kwargs):
+    """An internal function to be called in api-client when a websocket
+    connection is required for port forwarding. args and kwargs are the
+    parameters of apiClient.request method."""
+
+    query_params = kwargs.get("query_params")
+
+    ports = []
```
yliaog, line 453: so ports is a list, if the query_params is ports="80,80,80", ports would have three dups, better to remove dups up front

stream/ws_client.py
iciclespider: I 100% agree. I got confused by your confusion! After I implemented the local port/remote port handling, I realized that you did not understand that an actual local port was not used, and it finally dawned on me that this is really not necessary.

I'll back out that local/remote port work and address your other comments.

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+        initial frame contaning the port number that channel is associated with.
+        """
+
+        self.websocket = websocket
+        self.local_ports = {}
+        for ix, local_remote in enumerate(ports):
+            self.local_ports[local_remote[0]] = self._Port(ix, local_remote[1])
+        threading.Thread(
+            name="Kubernetes port forward proxy", target=self._proxy, daemon=True
+        ).start()
```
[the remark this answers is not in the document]
iciclespider: One such thread for each PortForward instance. It terminates after either side is closed once all pending data is flushed. I will comment it.

stream/ws_client.py
```diff
@@ -225,6 +227,143 @@ def close(self, **kwargs):
...
+                    data = port.python.recv(1024 * 1024)
+                    if data:
+                        kubernetes_data += ABNF.create_frame(
+                            port.channel + data,
+                            ABNF.OPCODE_BINARY,
+                        ).format()
+                    else:
+                        port.python.close()
+                        rlist.remove(s)
+                        if len(rlist) == 1:
```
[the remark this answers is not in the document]
iciclespider: I reworked the detection of how and when it is time to terminate this _proxy thread.

iciclespider:
@yliaog I submitted another commit for review.

stream/ws_client.py
```diff
@@ -285,42 +303,44 @@ def setsockopt(self, level, optname, value):
...
-                    wlist.append(port.python)
-            if kubernetes_data:
-                wlist.append(self.websocket.sock)
+                if port.python.fileno() != -1:
+                    if port.data:
+                        wlist.append(port.python)
+                        all_closed = False
+                    else:
+                        if self.websocket.connected:
+                            rlist.append(port.python)
```
yliaog, line 333: so currently rlist.append(port.python) is called only when both (port.data is None) and (self.websocket.connected is True)

IMO, whether port.data is None or not should not be used to decide whether to read from port.python, what do you think?

stream/ws_client.py
```diff
@@ -285,42 +303,44 @@ def setsockopt(self, level, optname, value):
...
+                            all_closed = False
+                        else:
+                            port.python.close()
+            if all_closed and (not self.websocket.connected or not kubernetes_data):
```
yliaog, line 337: if all_closed, and kubernetes_data, where is the 'kuberentes_data' going to be sent to?

stream/ws_client.py
iciclespider: Good catch, I noticed that myself and was already reworking this code.

stream/ws_client.py
iciclespider: `all_closed` is just all the local ports. I renamed this variable to be `local_all_closed`.

iciclespider (commented):
@yliaog I pushed a commit that addresses the comments.

stream/ws_client.py
yliaog: well, for the following condition:

 * all_closed, and self.websocket.connected,  and kubernetes_data *

the 'if' condition is line 337 would be false, hence, websokcet won't close, and the fuction won't return.

but since all_closed is true, even though kubernetes_data is not None, there is no socket to receive that kubernetes_data, there seems to be no point to keep the thread running.

yliaog: i think the scenario could happen when:

immediately after python application send some request, it closes the socket, instead of waiting to receive any response.

iciclespider: > but since all_closed is true, even though kubernetes_data is not None, there is no socket to receive that kubernetes_data, there seems to be no point to keep the thread running.

There is a socket to receive the data, the PortFoward.websocket.sock socket. That socket has not yet been closed by virtue of the fact that PortForward.websocket.connected is still True.

iciclespider: I reworked that if to instead be:
```
            if local_all_closed and not (self.websocket.connected and kubernetes_data):
```
I think it is clearer that way.

yliaog: ok.

stream/ws_client.py
```diff
@@ -225,6 +227,174 @@ def close(self, **kwargs):
...
+                else:
+                    port = local_ports[sock]
+                    data = port.python.recv(1024 * 1024)
+                    if data:
+                        kubernetes_data += ABNF.create_frame(
+                            port.channel + data,
+                            ABNF.OPCODE_BINARY,
+                        ).format()
+                    else:
+                        port.python.close()
```
yliaog, line 387: previously, there was the condition before the line above

                 if not port.data:

why is it removed?

iciclespider: I removed it because the socket has been closed by the python application. This also makes it symmetrical with the closing down of the websocket half. If there is `kubernetes_data` left over when the websocket becomes `not connected`, the `kubernetes_data` is not sent. Same with the python socket. There is no longer a functional socket to send the remaining data on.

stream/ws_client.py
```diff
@@ -225,6 +227,174 @@ def close(self, **kwargs):
...
+                    wlist.append(self.websocket)
+            local_all_closed = True
+            for port in self.local_ports.values():
+                if port.python.fileno() != -1:
+                    if port.error or not self.websocket.connected:
+                        if port.data:
+                            wlist.append(port.python)
+                            local_all_closed = False
+                        else:
+                            port.python.close()
```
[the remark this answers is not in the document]
iciclespider: > when there is port.error, as soon as port.data is flushed, port.python is closed (even when websocket.connected is true). i think that is correct, but want to double confirm.

That is intentional. The only time when a message arrives on the error channel is when the data channel either did not initialize or is not longer working.

> also port.error is not delivered to the python app, only port.data is delivered. it's worth to mention this behavior in the documentation.

There is the `PortForward.error(<port-number)` call to check if an error occurred.

stream/ws_client.py
yliaog: sounds good.

stream/ws_client.py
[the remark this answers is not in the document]
yliaog: yes, better to have it in the documentation. but i guess the examples / e2e tests would make it clear too.

yliaog:
thanks for the PR, really appreciate it.

/lgtm
/approve

PR #211, opened 2020-08-27
Refactor stream package to enable common method helpers for other streaming api classes.

iciclespider:
This puts in place the methods needed by this port forwarding implementation: #210 
Also, simplify the patten used to monkey patch the `ApiClient.request` method.

### DIFF
--- a/stream/stream.py
+++ b/stream/stream.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import types
+
 from . import ws_client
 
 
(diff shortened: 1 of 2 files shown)


stream/ws_client.py
```diff
@@ -259,41 +225,86 @@ def close(self, **kwargs):
...
+    # configuration
+    try:
+        return api_client.configuration
+    except AttributeError:
+        return api_client.config
+
+
+def websocket_call(api_client, _method, url, **kwargs):
     """An internal function to be called in api-client when a websocket
     connection is required. args and kwargs are the parameters of
```
yliaog, line 297: the comment needs to be updated

stream/stream.py
```diff
@@ -12,26 +12,19 @@
...
-            config = func.__self__.api_client.config
-
-        return ws_client.websocket_call(config, *args, **kwargs)
-
-    prev_request = func.__self__.api_client.request
+    api_client = func.__self__.api_client
+    prev_request = api_client.request
     try:
-        func.__self__.api_client.request = _intercept_request_call
+        api_client.request = types.MethodType(ws_client.websocket_call, api_client)
```
yliaog, line 27: why not using a function like before?

iciclespider: Good question. The original way, the ws_client.websocket_call only needed to know about the passed in `configuration`, and did not require knowing about api_client and now to get the configuration from that. I like that better.

I reworked this to obtain the `configuration` in stream.py, and then I used functools.partial rather then a function, to get the configuration to ws_client.websocket_call. Then, I enabled this logic to be shared with the eventual `portforward` method also using functools.partial. This means that the only change to stream.py for port forwarding will be the addition of this line:
```
portfoward = functools.partial(_websocket_reqeust, ws_client.portforward_call)
```

yliaog: i like it too.

yliaog:
thanks for the pr
/lgtm
/approve

PR #214, opened 2020-10-14
do-not-merge: run dynamic client e2e test and verify the exception

roycaihw:
make sure the same test failure gets caught: https://github.com/kubernetes-client/python/pull/1281

Note that the failure is flaky, since the client won't raise exception if the discovery is already updated. It only fails when a retry is needed. 

/hold


roycaihw:
the same error is caught: https://travis-ci.org/github/kubernetes-client/python-base/builds/735584773

PR #221, opened 2020-12-25
Add load_config method which allows a more generic way to load the kubeconfig

MoShitrit:
This is the first out of two changes in relation to https://github.com/kubernetes-client/python/issues/1005.
In this PR, I'm adding a `load_config()` method which applies a logic quite similar to how the [Go client operates](https://github.com/kubernetes/client-go/blob/9346d688810d23f933041c41e8650967d639be76/tools/clientcmd/client_config.go#L536-L553).
I've applied it following [this](https://github.com/kubernetes-client/python/issues/1005#issuecomment-587969252) suggestion.
Once this is merged and approved. I'll have another PR to the `kubernetes-client/python` repo to update the examples to use this new method generally.

Side node:
I also made a minor change to the `load_kube_config` module and function to assign the value of `KUBE_CONFIG_DEFAULT_LOCATION` constant in a more "pythonic" way of using `pathlib.Path.home()` instead of hardcoding `~` as a reference to the homedir.

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -13,14 +13,12 @@
 # limitations under the License.
 
 import os
+
 from .config_exception import ConfigException
 from .incluster_config import load_incluster_config
-from .kube_config import (
-    list_kube_config_contexts,
-    load_kube_config,
-    load_kube_config_from_dict,
-    new_client_from_config,
-    KUBE_CONFIG_DEFAULT_LOCATION)
+from .kube_config import (KUBE_CONFIG_DEFAULT_LOCATION,
+                          list_kube_config_contexts, load_kube_config,
+                          load_kube_config_from_dict, new_client_from_config)
 
 
 def load_config(**kwargs):
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -26,10 +26,13 @@ from .kube_config import (
 def load_config(**kwargs):
     """
     Wrapper function to load the kube_config.
-    It will initially try to load_kube_config from provided path, then check if the KUBE_CONFIG_DEFAULT_LOCATION exists
-    If neither exists- it will fall back to load_incluster_config and inform the user accordingly.
+    It will initially try to load_kube_config from provided path,
+    then check if the KUBE_CONFIG_DEFAULT_LOCATION exists
+    If neither exists- it will fall back to load_incluster_config
+    and inform the user accordingly.
 
-    :param kwargs: A combination of all possible kwargs that can be passed to either load_kube_config or
+    :param kwargs: A combination of all possible kwargs that
+    can be passed to either load_kube_config or
     load_incluster_config functions.
     """
     if "kube_config_path" in kwargs.keys() or os.path.exists(
@@ -37,6 +40,8 @@ def load_config(**kwargs):
         load_kube_config(**kwargs)
     else:
         print(
-            "kube_config_path not provided and default location ({0}) does not exist. "
-            "Using inCluster Config. This might not work.".format(KUBE_CONFIG_DEFAULT_LOCATION))
+            "kube_config_path not provided and "
+            "default location ({0}) does not exist. "
+            "Using inCluster Config. "
+            "This might not work.".format(KUBE_CONFIG_DEFAULT_LOCATION))
         load_incluster_config(**kwargs)
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -12,7 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 from .config_exception import ConfigException
 from .incluster_config import load_incluster_config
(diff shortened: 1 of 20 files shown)


roycaihw:
@MoShitrit There is some test failure (invalid syntax) in https://travis-ci.org/github/kubernetes-client/python-base/jobs/751491549. Could you take a look?

MoShitrit:
> @MoShitrit There is some test failure (invalid syntax) in https://travis-ci.org/github/kubernetes-client/python-base/jobs/751491549. Could you take a look?

Hey @roycaihw 
That invalid syntax is probably because of the [switch I made](https://github.com/kubernetes-client/python-base/pull/221/files#diff-aff290ec1059ab1072bac2c2345dcf920b3535286175ac36d80dfcca5493be6cL48-R49) from the hardcoded `~/.kube/config` path to use the more Pythonic `pathlib` library, which only lives in Python3 (The error appears to be on a test with Python 2.7).
Given that Python2 is at EOL for over a year now- should it still be supported?

yliaog:
to drop support for python 2, we need to send an email to dev to announce it, and make it clear from which release python 2 is no longer supported.

MoShitrit:
> to drop support for python 2, we need to send an email to dev to announce it, and make it clear from which release python 2 is no longer supported.

@yliaog gotcha, yeah that makes sense.
OK I'll revert this change then, which should probably fix these test failures. 
Thanks!

config/__init__.py
```diff
@@ -12,7 +12,25 @@
...
 
+import os
 from .config_exception import ConfigException
 from .incluster_config import load_incluster_config
 from .kube_config import (list_kube_config_contexts, load_kube_config,
-                          load_kube_config_from_dict, new_client_from_config)
+                          load_kube_config_from_dict, new_client_from_config, KUBE_CONFIG_DEFAULT_LOCATION)
+
+
+def load_config(**kwargs):
```
roycaihw, line 22: perhaps the import failure is related to this function. Why do we put this function in __init__.py?

MoShitrit: I think I just realized what the issue is.. 
image

The support for `fstrings` was only introduced in Python 3.6, which is why all the tests for versions prior to 3.6 are failing 🤦🏼 
I'll switch this `print` statement to use the legacy `.format()` method, that should fix it. Giving that a try..

MoShitrit: Yep, looks like that did it! All the functional tests are now passing.
I seem to have one last failure now with `update-pycodestyle` which is complaining about the line length in two files, so I'm going to lint them using `black` and give that another go.

MoShitrit: @roycaihw tests are now passing! 🎉

MoShitrit:
Hey @roycaihw 
Still looking to get this one approved when you get a chance please 😄 
Thanks!

config/__init__.py
```diff
@@ -12,7 +12,34 @@
...
+                          list_kube_config_contexts, load_kube_config,
                           load_kube_config_from_dict, new_client_from_config)
+
+
+def load_config(**kwargs):
+    """
+    Wrapper function to load the kube_config.
+    It will initially try to load_kube_config from provided path,
+    then check if the KUBE_CONFIG_DEFAULT_LOCATION exists
+    If neither exists- it will fall back to load_incluster_config
```
yliaog, line 29: s/exists-/exists,/

yliaog (commented):
it looks good, could you please add a test case?

MoShitrit:
> it looks good, could you please add a test case?

Yep! Sure can.
I'll work on that soon and add a test case.
Thanks!

MoShitrit:
> > it looks good, could you please add a test case?
> 
> Yep! Sure can.
> I'll work on that soon and add a test case.
> Thanks!

@yliaog Where should I put that test case? Should it live under the same repo through that PR? ([here](https://github.com/kubernetes-client/python-base/tree/master/dynamic)?) Or should I add it to the main python repo through a different PR? ([here](https://github.com/kubernetes-client/python/tree/master/kubernetes/test)?)

config/__init__.py
MoShitrit: Changed via https://github.com/kubernetes-client/python-base/pull/221/commits/6d1c8d3713057e87d973d853b36373c06901d092

roycaihw:
/lgtm
/approve

There are two parts need to be tested:
1. out-of-cluster kubeconfig. I can think of: 1) we either change [get_e2e_configuration](https://github.com/kubernetes-client/python/blob/master/kubernetes/e2e_test/base.py) to use `load_config`, or 2) add a separate `get_e2e_configuration_load_config` method and add a [test](https://github.com/kubernetes-client/python/blob/master/kubernetes/e2e_test/test_client.py) that uses it. Please send a PR to the main repo after this PR is merged.
2. in-cluster kubeconfig. This requires creating a pod that runs the Python client.

PR #226, opened 2021-01-29
fix: load cache error when CacheDecoder object is not callable

jamesgetx:
fix the code snippet `self._cache = json.load(f, cls=CacheDecoder(self.client))` raise `'CacheDecoder' object is not callable`


yliaog:
thanks for the pr, could you please add a test case?

jamesgetx:
> thanks for the pr, could you please add a test case?

Ok, I will add it later:)

jamesgetx:
hi @yliaog. I have pushed the test case, but i'm not sure there is any other better solutions to test the code `self._cache = json.load(f, cls=partial(CacheDecoder, self.client))`

dynamic/test_discovery.py
```diff
@@ -0,0 +1,40 @@
...
+from . import DynamicClient
+
+
+class TestDiscoverer(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.config = base.get_e2e_configuration()
+
+    def test_init_cache_from_file(self):
```
yliaog, line 30: so if you run the test without the fix, the test will fail, and there will be exception, right?

jamesgetx: yes. without the fix, `_Discoverer__init_cache` will raise `'CacheDecoder' object is not callable` then call `self.invalidate_cache()`,  finally `self._write_cache()` will be called to modify `_Discoverer__cache_file`

PR #227, opened 2021-02-22
Fix bug with Watch and 410 retries

chrisayoub:
There is a bug that was introduced with the following PR:
https://github.com/kubernetes-client/python-base/pull/133

When there is a 410 error that needs to be retried and the user specifics any timeout values (`timeouts`), the code currently returns, when it really should proceed to the next iteration of the for loop and actually do the retry.

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -151,7 +151,9 @@ class Watch(object):
         if 'resource_version' in kwargs:
             self.resource_version = kwargs['resource_version']
 
-        timeouts = ('timeout_seconds' in kwargs)
+        # Do not attempt retries if user specifies a timeout.
+        # We want to ensure we are returning within that timeout.
+        disable_retries = ('timeout_seconds' in kwargs)
         retry_after_410 = False
         while True:
             resp = func(*args, **kwargs)
@@ -164,9 +166,9 @@ class Watch(object):
                         if isinstance(event, dict) \
                                 and event['type'] == 'ERROR':
                             obj = event['raw_object']
-                            # Current request expired, let's retry,
+                            # Current request expired, let's retry, (if enabled)
                             # but only if we have not already retried.
-                            if not retry_after_410 and \
+                            if not disable_retries and not retry_after_410 and \
                                     obj['code'] == HTTP_STATUS_GONE:
                                 retry_after_410 = True
                                 break
@@ -190,5 +192,5 @@ class Watch(object):
                 else:
                     self._stop = True
 
-            if timeouts or self._stop:
+            if self._stop or disable_retries:
                 break
--- a/watch/watch_test.py
+++ b/watch/watch_test.py
@@ -288,14 +288,64 @@ class WatchTests(unittest.TestCase):
         fake_api.get_thing = Mock(return_value=fake_resp)
 
         w = Watch()
+        # No events are generated when no initial resourceVersion is passed
+        # No retry is attempted either, preventing an ApiException
+        assert not list(w.stream(fake_api.get_thing))
+
+        fake_api.get_thing.assert_called_once_with(
+            _preload_content=False, watch=True)
+        fake_resp.read_chunked.assert_called_once_with(decode_content=False)
+        fake_resp.close.assert_called_once()
+        fake_resp.release_conn.assert_called_once()
+
+    def test_watch_retries_on_error_event(self):
+        fake_resp = Mock()
+        fake_resp.close = Mock()
+        fake_resp.release_conn = Mock()
+        fake_resp.read_chunked = Mock(
+            return_value=[
+                '{"type": "ERROR", "object": {"code": 410, '
+                '"reason": "Gone", "message": "error message"}}\n'])
+
+        fake_api = Mock()
+        fake_api.get_thing = Mock(return_value=fake_resp)
+
+        w = Watch()
         try:
-            for _ in w.stream(fake_api.get_thing):
+            for _ in w.stream(fake_api.get_thing, resource_version=0):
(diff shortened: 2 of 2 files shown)


yliaog:
if timeout or stop, retry should *not* be attempted? i think the current behavior is expected.

PaulFurtado:
@yliaog If you look at the `timeouts` variable in the code, it is an argument to this function for a user-specified watch timeout for the request, it does not indicate that a timeout has actually occurred.

So for clients that specify a timeout as an argument to the function, the result of this is that when a 410 error occurs, the caller gets no exception at all and it appears that the watch has gracefully ended, which makes it impossible for a caller to recover from 410.

yliaog:
i see. in that case, timeouts condition is not useful. but it should still break without retry when stop is true. so what about the following:

if self._stop:
  break

chrisayoub:
This is the previous PR and issue where this `timeouts` functionality was introduced:
https://github.com/kubernetes-client/python-base/pull/36
https://github.com/kubernetes-client/python/issues/124

Do you think that change would conflict with this?

PaulFurtado:
Yeah, removing the timeouts condition would break functionality. If that were set, a user might end up having a watch which lasts forever despite having set a timeout on it explicitly.

Thinking about this more, I actually think the best fix would be to never do any type of retry when `timeouts` is set because then it is impossible to guarantee that the function exits within the specified timeout. That would remain true to the way that this code worked prior to the 410 error handling. To achieve that, line 169 could be altered to include `timeouts is not None` so that we don't retry 410 when the timeouts parameter is set.

Additionally, I think maybe we should make that intention more clear so that similar bugs don't get introduced in the future. Before the while loop, we could set a variable like:
```python
disable_retries = timeouts is not None
```
and then alter this condition to be:
```python
if self._stop or disable_retries:
    break
```
and change the condition on line 169 to:
```python
if not disable_retries and not retry_after_410 and obj['code'] == HTTP_STATUS_GONE:
```

chrisayoub:
@PaulFurtado I updated the PR with your suggested changes, and I tested and they seem to work correctly for me.

@yliaog can you take another look? Thanks!

yliaog:
i'm not sure how the PR solve the original problem as given in the PR description below, so when there is 410 error, and timeout is given, there will be no retry at all, it returns.

 "When there is a 410 error that needs to be retried and the user specifics any timeout values (timeouts), the code currently returns, when it really should proceed to the next iteration of the for loop and actually do the retry."

chrisayoub:
I think this is just a question of the desired behavior ultimately. When the user specifies a timeout value and we encounter an event of type `error`, should we yield that event as normal, or should we raise an `ApiException`? I can easily modify the PR to do whichever seems more correct here.

Ultimately, the desired fix is that under these circumstances, we do not simply return from the function, and we either yield the event or raise an Exception.

PaulFurtado:
@yliaog by adding the `not disable_retries` condition to the if statement on line 169, it hits the `else` block of that if statement which immediately raises the ApiException indicating that the 410 occurred, so that the caller can deal with the 410 error.

watch/watch.py
```diff
@@ -166,7 +166,7 @@ def stream(self, func, *args, **kwargs):
                             obj = event['raw_object']
                             # Current request expired, let's retry,
                             # but only if we have not already retried.
-                            if not retry_after_410 and \
```
yliaog, line 169: please add a comment about the added condition "not disable_retries and not"

yliaog:
looks good to disable retry for the 410 case, could you please add a test case for it?

watch/watch_test.py
```diff
@@ -287,15 +287,65 @@ def test_watch_with_error_event(self):
         fake_api = Mock()
         fake_api.get_thing = Mock(return_value=fake_resp)
 
+        w = Watch()
+        # No events are generated when no initial resourceVersion is passed
+        # No retry is attempted either, preventing an ApiException
+        assert not list(w.stream(fake_api.get_thing))
```
chrisayoub, line 293: I updated the existing test case, which was not actually correct for the current version of the code with retries. Additionally, I added 2 more test cases to test both code paths that my PR is related to.

watch/watch.py
chrisayoub: Done, see lines 154-155 as well

watch/watch_test.py
```diff
@@ -287,15 +287,65 @@ def test_watch_with_error_event(self):
...
+            return_value=[
+                '{"type": "ERROR", "object": {"code": 410, '
+                '"reason": "Gone", "message": "error message"}}\n'])
+
+        fake_api = Mock()
+        fake_api.get_thing = Mock(return_value=fake_resp)
+
+        w = Watch()
+        try:
+            for _ in w.stream(fake_api.get_thing, timeout_seconds=10):
```
yliaog, line 342: no resource_version?

chrisayoub: For testing this code path and condition, you do not actually need to supply `resource_version` here, as an exception is raised and only the code in the watch `finally` block will execute.

yliaog: ok

yliaog:
thanks for the pr.

/lgtm
/approve

chrisayoub:
@yliaog it looks like the failed CI run is preventing this from automatically merging:
https://travis-ci.org/github/kubernetes-client/python-base/jobs/760405517

However, this test case isn't even in this repo? Am I supposed to do anything else to proceed further in getting this merged? 

Thank you!

yliaog:
closing the pr, and reoopen to trigger another CI run

PR #231, opened 2021-04-08
Fixes kubernetes-client/python issue 1047 "ResponseNotChunked from watch"

dhague:
In recent versions of K8S (>1.16?), when a `Watch.stream()` call uses a
resource_version which is too old the resulting 410 error is wrapped in JSON
and returned in a non-chunked 200 response. Using `resp.stream()` instead of
`resp.read_chunked()` automatically handles the response being either chunked or
non-chunked.

### DIFF
--- a/watch/watch.py
+++ b/watch/watch.py
@@ -53,7 +53,7 @@ def _find_return_type(func):
 
 def iter_resp_lines(resp):
     prev = ""
-    for seg in resp.read_chunked(decode_content=False):
+    for seg in resp.stream(amt=None, decode_content=False):
         if isinstance(seg, bytes):
             seg = seg.decode('utf8')
         seg = prev + seg
--- a/watch/watch_test.py
+++ b/watch/watch_test.py
@@ -30,7 +30,7 @@ class WatchTests(unittest.TestCase):
         fake_resp = Mock()
         fake_resp.close = Mock()
         fake_resp.release_conn = Mock()
-        fake_resp.read_chunked = Mock(
+        fake_resp.stream = Mock(
             return_value=[
                 '{"type": "ADDED", "object": {"metadata": {"name": "test1",'
                 '"resourceVersion": "1"}, "spec": {}, "status": {}}}\n',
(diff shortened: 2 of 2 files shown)


watch/watch.py
```diff
@@ -53,7 +53,7 @@ def _find_return_type(func):
 
 def iter_resp_lines(resp):
     prev = ""
-    for seg in resp.read_chunked(decode_content=False):
+    for seg in resp.stream(decode_content=False):
```
yliaog, line 56: https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html

stream(amt=65536, decode_content=None)¶

the default amt it 65536, that does not seem desirable.

dhague: Good catch, thanks - fix incoming...

yliaog:
please squash the commits

have you manually tested it in your env to verify the fix? 

The documentation below says stream is a wrapper for read(), but does not mention read_chunked(). 

stream(amt=65536, decode_content=None)
A generator wrapper for the read() method.

dhague:
1. Commits are now squashed

2. I have been testing this in my dev environment, and also in the CI pipeline for our app that uses the Python Kubernetes client. Some flaky tests in the CI pipeline are what led me to fixing this in the first place :-)

3. As to the `stream()` method itself, you can see from the implementation below that it deals explicitly with both `read()` and `read_chunked()` cases - it does exactly what we want:
```
    def stream(self, amt=2 ** 16, decode_content=None):
        """
        A generator wrapper for the read() method. A call will block until
        ``amt`` bytes have been read from the connection or until the
        connection is closed.

        :param amt:
            How much of the content to read. The generator will return up to
            much data per iteration, but may return less. This is particularly
            likely when using compressed data. However, the empty string will
            never be returned.

        :param decode_content:
            If True, will attempt to decode the body based on the
            'content-encoding' header.
        """
        if self.chunked and self.supports_chunked_reads():
            for line in self.read_chunked(amt, decode_content=decode_content):
                yield line
        else:
            while not is_fp_closed(self._fp):
                data = self.read(amt=amt, decode_content=decode_content)

                if data:
                    yield data
```

yliaog:
thanks for pr

/lgtm
/approve

PR #233, opened 2021-04-12
load_kube_config_from_dict() support define custom temp files path

onecer:
Centos '/tmp' will clear files that created over 10 days ago in this directory. 

So, If we can define a temp files path then problems will be solved.

It works for [k8s config tmp files cleared by centos](https://github.com/kubernetes-client/python/issues/765)
and [SSLError: FileNotFoundError: [Errno 2] No such file or directory](https://github.com/kubernetes-client/python/issues/1236)

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -60,7 +60,7 @@ def _cleanup_temp_files():
     _temp_files = {}
 
 
-def _create_temp_file_with_content(content):
+def _create_temp_file_with_content(content, temp_file_path=None):
     if len(_temp_files) == 0:
         atexit.register(_cleanup_temp_files)
     # Because we may change context several times, try to remember files we
@@ -68,7 +68,9 @@ def _create_temp_file_with_content(content):
     content_key = str(content)
     if content_key in _temp_files:
         return _temp_files[content_key]
-    _, name = tempfile.mkstemp()
+    if temp_file_path and not os.path.isdir(temp_file_path):
+        os.makedirs(name=temp_file_path)
+    _, name = tempfile.mkstemp(dir=temp_file_path)
     _temp_files[content_key] = name
     with open(name, 'wb') as fd:
         fd.write(content.encode() if isinstance(content, str) else content)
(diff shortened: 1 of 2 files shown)


yliaog:
please add a test

onecer:
> please add a test

done.

config/kube_config_test.py
```diff
@@ -1290,6 +1290,21 @@ def test_load_kube_config_from_dict(self):
...
+        tmp_path = os.path.join(
+            os.path.dirname(
+                os.path.dirname(
+                    os.path.abspath(__file__))),
+            'tmp_file_path_test')
+        load_kube_config_from_dict(config_dict=self.TEST_KUBE_CONFIG,
+                                   context="simple_token",
+                                   client_configuration=actual,
+                                   temp_file_path=tmp_path)
+        self.assertEqual(expected, actual)
```
yliaog, line 1306: add assertion that the tmp_path is used for creating the temp file?

onecer: I have been update the assertion.

with context="ssl"  certificate will be store in tmp_path . 

`self.assertFalse(True if not os.listdir(tmp_path) else False)`

config/kube_config_test.py
```diff
@@ -1290,6 +1290,20 @@ def test_load_kube_config_from_dict(self):
...
+        tmp_path = os.path.join(
+            os.path.dirname(
+                os.path.dirname(
+                    os.path.abspath(__file__))),
+            'tmp_file_path_test')
+        load_kube_config_from_dict(config_dict=self.TEST_KUBE_CONFIG,
+                                   context="ssl",
+                                   client_configuration=actual,
+                                   temp_file_path=tmp_path)
+        self.assertFalse(True if not os.listdir(tmp_path) else False)
```
yliaog, line 1304: self.assertEqual(expected, actual) is also useful, could you add it back?

config/kube_config.py
```diff
@@ -69,7 +69,7 @@ def _create_temp_file_with_content(content, temp_file_path=None):
     if content_key in _temp_files:
         return _temp_files[content_key]
     if temp_file_path and not os.path.isdir(temp_file_path):
-        os.makedirs(name=temp_file_path, exist_ok=True)
```
yliaog: why is exist_ok=True removed?

onecer: `not os.path.isdir(temp_file_path)` have been consider is directory not exists situation.

yliaog:
thanks for the pr. please squash the commits.
/lgtm

config/kube_config_test.py
onecer: have been add it back

PR #234, opened 2021-04-13
quick fix of decoding error for BOOKMARK event

yliaog:
Fixes https://github.com/kubernetes-client/python/issues/1404

### DIFF
--- a/watch/watch_test.py
+++ b/watch/watch_test.py
@@ -255,6 +255,19 @@ class WatchTests(unittest.TestCase):
         self.assertEqual("1", event['object']['metadata']['resourceVersion'])
         self.assertEqual("1", w.resource_version)
 
+    def test_unmarshal_with_bookmark(self):
+        w = Watch()
+        event = w.unmarshal_event(
+            '{"type":"BOOKMARK","object":{"kind":"Job","apiVersion":"batch/v1"'
+            ',"metadata":{"resourceVersion":"1"},"spec":{"template":{'
+            '"metadata":{},"spec":{"containers":null}}},"status":{}}}',
+            'V1Job')
+        self.assertEqual("BOOKMARK", event['type'])
+        # Watch.resource_version is *not* updated, as BOOKMARK is treated the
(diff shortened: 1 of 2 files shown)


watch/watch_test.py
```diff
@@ -255,6 +255,19 @@ def test_unmarshal_with_custom_object(self):
         self.assertEqual("1", event['object']['metadata']['resourceVersion'])
         self.assertEqual("1", w.resource_version)
 
+    def test_unmarshal_with_bookmark(self):
+        w = Watch()
+        event = w.unmarshal_event(
+            '{"type":"BOOKMARK","object":{"kind":"Job","apiVersion":"batch/v1"'
+            ',"metadata":{"resourceVersion":"1"},"spec":{"template":{'
+            '"metadata":{},"spec":{"containers":null}}},"status":{}}}',
```
roycaihw, line 263: nit: reading the [doc](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks), it looks like a bookmark event won't contain things like job spec and job status.

roycaihw:
/lgtm

https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks

I'm not familiar with this feature and don't know how it's supposed to be used. If we find a use case in this client maybe we should open a feature request to support it.

watch/watch_test.py
yliaog: that's correct, the spec and status are all empty, even though they seem to contain something.

roycaihw: I thought the key would be omitted, but maybe that's the wrong impression the doc gave me

yliaog: it is not. kubernetes-client/python#1404 has the BOOKMARK event:

{"type":"BOOKMARK","object":{"kind":"Job","apiVersion":"batch/v1","metadata":{"resourceVersion":"45398385","creationTimestamp":null},"spec":{"template":{"metadata":{"creationTimestamp":null},"spec":{"containers":null}}},"status":{}}}

PR #236, opened 2021-04-27
Support customizing “Accept” header #1428

Yashks1994:
####

We are adding the new header parameter in the dynamic client which helps in scenarios where the user only cares about object metadata and wants to reduce the client-side deserialisation time.

This changes will accept the header like `'application/json;as=PartialObjectMetadata;v=v1;g=meta.k8s.io'` or `'application/json;as=Table;v=v1;g=meta.k8s.io'`.

Fixes #1428

#### Special notes for your reviewer:
Added the metadata field which is currently present in [client-go](https://github.com/kubernetes/client-go/blob/master/metadata/metadata.go#L162). Now this feature will also be supporting in the python-client.

```release-note
The dynamic client now supports customizing http "Accept" header through the `header_params` parameter, which can be used to customizing API server response, e.g. retrieving object metadata only. 
```

```docs

```

### DIFF
--- a/dynamic/client.py
+++ b/dynamic/client.py
@@ -219,11 +219,14 @@ class DynamicClient(object):
         header_params = params.get('header_params', {})
         form_params = []
         local_var_files = {}
-        # HTTP header `Accept`
-        header_params['Accept'] = self.client.select_header_accept([
-            'application/json',
-            'application/yaml',
-        ])
+
+        # Checking Accept header.
+        new_header_params = dict((key.lower(), value) for key, value in header_params.items())
+        if not 'accept' in new_header_params:
+            header_params['Accept'] = self.client.select_header_accept([
+                'application/json',
+                'application/yaml',
+            ])
 
         # HTTP header `Content-Type`
         if params.get('content_type'):
--- a/dynamic/test_client.py
+++ b/dynamic/test_client.py
@@ -359,7 +359,7 @@ class TestDynamicClient(unittest.TestCase):
 
         resp = api.get(namespace='default', pretty=True, label_selector="e2e-test=true")
         self.assertEqual([], resp.items)
-
+    
     def test_node_apis(self):
         client = DynamicClient(api_client.ApiClient(configuration=self.config))
         api = client.resources.get(api_version='v1', kind='Node')
@@ -367,3 +367,19 @@ class TestDynamicClient(unittest.TestCase):
         for item in api.get().items:
             node = api.get(name=item.metadata.name)
             self.assertTrue(len(dict(node.metadata.labels)) > 0)
+    
+    # test_node_apis_partial_object_metadata lists all nodes in the cluster, but only retrieves object metadata
+    def test_node_apis_partial_object_metadata(self):
+        client = DynamicClient(api_client.ApiClient(configuration=self.config))
+        api = client.resources.get(api_version='v1', kind='Node')
+        
+        params = {'header_params': {'Accept': 'application/json;as=PartialObjectMetadataList;v=v1;g=meta.k8s.io'}}
+        resp = api.get(**params)
+        self.assertEqual('PartialObjectMetadataList', resp.kind)
+        self.assertEqual('meta.k8s.io/v1', resp.apiVersion)
+
+        params = {'header_params': {'aCcePt': 'application/json;as=PartialObjectMetadataList;v=v1;g=meta.k8s.io'}}
+        resp = api.get(**params)
+        self.assertEqual('PartialObjectMetadataList', resp.kind)
+        self.assertEqual('meta.k8s.io/v1', resp.apiVersion)
+


dynamic/client.py
```diff
@@ -224,6 +224,11 @@ def request(self, method, path, body=None, **params):
             'application/json',
             'application/yaml',
         ])
+        # HTTP header `application/vnd.kubernetes.protobuf;as=PartialObjectMetadata;g=meta.k8s.io;v=v1,application/json;as=PartialObjectMetadata;g=meta.k8s.io;v=v1,application/json`
+        header_params['application/vnd.kubernetes.protobuf;as=PartialObjectMetadata;g=meta.k8s.io;v=v1,application/json;as=PartialObjectMetadata;g=meta.k8s.io;v=v1,application/json'] = self.client.select_header_accept([
+            'application/json',
+            'application/yaml',
+        ])
```
yliaog, line 231: please add a test to verify it indeed gets the metadata only

Yashks1994: I have added a new condition to check if header is present or not.

dynamic/client.py
```diff
@@ -216,14 +216,22 @@ def request(self, method, path, body=None, **params):
...
             'application/json',
+            ])
+        if params.get('application/yaml') is not None:
+            header_params['Accept'] = self.client.select_header_accept([
             'application/yaml',
-        ])
+            ])
+        if params.get('application/vnd.kubernetes.protobuf') is not None:
+            header_params['Accept'] = self.client.select_header_accept([
+            'application/vnd.kubernetes.protobuf',
```
roycaihw, line 231: the Python client doesn't understand protobuf. What we want is `g=meta.k8s.io;v=v1,application/json;as=PartialObjectMetadata `

yliaog: that's right. it's also why it's good to have a test to verify it works as expected

dynamic/client.py
```diff
@@ -219,11 +219,16 @@ def request(self, method, path, body=None, **params):
         header_params = params.get('header_params', {})
         form_params = []
         local_var_files = {}
+
         # HTTP header `Accept`
-        header_params['Accept'] = self.client.select_header_accept([
-            'application/json',
-            'application/yaml',
-        ])
+        if params.get('header_params') is not None:
```
yliaog, line 224: could use header_params directly which is assigned at line 219

dynamic/client.py
```diff
@@ -219,11 +219,16 @@ def request(self, method, path, body=None, **params):
...
+            partial_object = params.get('header_params')
```
yliaog, line 225: ditto, header_params can be used, instead of partial_object here

dynamic/client.py
```diff
@@ -219,11 +219,16 @@ def request(self, method, path, body=None, **params):
...
+            if 'PartialObjectMetadata' in partial_object.get('Accept'):
+                header_params['Accept'] = 'application/json;as=PartialObjectMetadata;v=v1;g=meta.k8s.io'
```
yliaog, line 227: better to just take whatever is passed, instead of overwriting it here, i think. so if header_params['Accept'] is present, then do nothing, otherwise, set the default ['application/json', 'application/yaml']

roycaihw: +1 to what Yu said. Let's make the dynamic client dynamic. I'm interested to see how this feature will work with things like `as=Table` (what kubectl uses)

roycaihw:
After this is merged and included in the main repo, we should have an example for this feature. 

/kind feature

dynamic/test_client.py
```diff
@@ -77,6 +77,17 @@ def test_cluster_custom_resources(self):
...
+        #Checking header_params `PartialObjectMetadata`
+        header_params = 'PartialObjectMetadata'
+        params = {'header_params': {'Accept': 'application/json;as=PartialObjectMetadata;v=v1;g=meta.k8s.io'}}
+        api_version = 'meta.k8s.io/v1'
+        resp = crd_api.get(
+            name=name,
+            **params
+        )
+        self.assertEqual(header_params, resp.kind)
+        self.assertEqual(api_version, resp.apiVersion)
```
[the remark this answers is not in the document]
Yashks1994: I have added a standalone testcase like you mentioned but it has some minor change.
I am passing name also as paramaters then only it is working `resp = api.get(name, **params)`.

But I simply pass this `resp = api.get(**params)` as to get the resp, it is throwing the below error:
```Traceback (most recent call last):
  File "/Users/yashkumarsingh/Desktop/OpenSource/python-base/tt.py", line 18, in <module>
    resp = api.get(**params)

  File "/Users/yashkumarsingh/Desktop/OpenSource/python-base/dynamic/client.py", line 109, in get
    return self.request('get', path, **kwargs)

  File "/Users/yashkumarsingh/Desktop/OpenSource/python-base/dynamic/client.py", line 58, in inner
    raise api_exception(e)
  dynamic.exceptions.DynamicApiError: 406
  Reason: Not Acceptable
  HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Date': 'Sat, 01 May 2021 04:31:15 GMT', 'Content-Length': '234'})
HTTP response body: b'{\n  "kind": "Status",\n  "apiVersion": "v1",\n  "metadata": {},\n  "status": "Failure",\n  "message": "you requested PartialObjectMetadata, but the requested object is a list (*core.NodeList)",\n  "reason": "NotAcceptable",\n  "code": 406\n}'
Original traceback: 
  File "/Users/yashkumarsingh/Desktop/OpenSource/python-base/dynamic/client.py", line 56, in inner
    resp = func(self, *args, **kwargs)

File: dynamic/client.py (hunk 3 above)
Comment (Yashks1994) [reply dynamic/client.py]: Thank for the suggestion, I have updated the code.

File: dynamic/client.py [hunk 7]
```diff
@@ -219,11 +219,15 @@ def request(self, method, path, body=None, **params):
...
         local_var_files = {}
+
         # HTTP header `Accept`
-        header_params['Accept'] = self.client.select_header_accept([
-            'application/json',
-            'application/yaml',
-        ])
+        if params.get('header_params'):
+            header_params = params.get('header_params')
+            header_params['Accept'] = header_params.get('Accept')
```
yliaog, line 226: why is the line above needed? i think you want to check whether header_params has 'Accept'

Yashks1994: If you check the test case there I am mentioned the type of header it will accept.

Through this only we can actually extract the value of params and pass it on to header_params[’Accept’]

yliaog: what about adding a negative test case? i.e. the header_params does *not* have 'Accept', check that the response is *not* particalobjectmeta

Yashks1994: Of doesn't have Accept and header with the value then we can directly substitute the value.

If you're recommending if that condition is true then we need another elif statement which will accept that condition

roycaihw: instead of asking the user to pass in `{'header_params': {'Accept': 'application/json;as=PartialObjectMetadata;v=v1;g=meta.k8s.io'}}`, why don't we just ask for `'application/json;as=PartialObjectMetadata;v=v1;g=meta.k8s.io'`? What's the benefit the other way? It adds more failure patterns and it's a worse user experience

Yashks1994: Ok 
Then I will update it like this. So user can directly pass it as simple dictionary.

dynamic/test_client.py
```diff
@@ -359,11 +359,27 @@ def test_configmap_apis(self):
...
         for item in api.get().items:
             node = api.get(name=item.metadata.name)
             self.assertTrue(len(dict(node.metadata.labels)) > 0)
+    
+    # test_node_apis_partial_object_metadata lists all nodes in the cluster, but only retrieves object metadata
+    def test_node_apis_partial_object_metadata(self):
+        client = DynamicClient(api_client.ApiClient(configuration=self.config))
+        api = client.resources.get(api_version='v1', kind='Node')
+
+        for i in api.get().items:
```
roycaihw, line 376: > test_node_apis_partial_object_metadata lists all nodes in the cluster, but only retrieves object metadata

The comment suggests using the partial object metadata feature in this **List** call, instead of doing additional **Get** calls below. A known use case of partial object metadata is to reduce the size of the API server response when you list objects in the cluster

dynamic/test_client.py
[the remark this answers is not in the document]
roycaihw: > "message": "you requested PartialObjectMetadata, but the requested object is a list (*core.NodeList)",\n  "reason": "NotAcceptable",\n  "code": 406

could you try replacing `PartialObjectMetadata` with `PartialObjectMetadataList`?

dynamic/client.py
Yashks1994: I am passing it direct assignment. It is giving the error.

`  File "/Users/yashkumarsingh/Desktop/OpenSource/python-base/dynamic/client.py", line 109, in get
    return self.request('get', path, **kwargs)
  File "/Users/yashkumarsingh/Desktop/OpenSource/python-base/dynamic/client.py", line 56, in inner
    resp = func(self, *args, **kwargs)
  File "/Users/yashkumarsingh/Desktop/OpenSource/python-base/dynamic/client.py", line 258, in request
    header_params['Accept'] = params['header_params']
TypeError: 'str' object does not support item assignment`

dynamic/test_client.py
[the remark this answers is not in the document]
Yashks1994: It worked.

dynamic/client.py
```diff
@@ -219,11 +219,15 @@ def request(self, method, path, body=None, **params):
         header_params = params.get('header_params', {})
```
roycaihw, line 219: sorry I didn't realize we had a `header_params` already. It looks like the user can already pass in 'Accept' but it just gets overwritten unconditionally. 

let's do this then:
```
# User didn't specify the Accept header. Default it to json and yaml.
if not header_params.has_key('Accept'):
    header_params['Accept'] = self.client.select_header_accept([
        'application/json',
        'application/yaml',
    ])
```

Yashks1994: Done.

dynamic/client.py
```diff
@@ -219,8 +219,10 @@ def request(self, method, path, body=None, **params):
...
         form_params = []
         local_var_files = {}
-        # HTTP header `Accept`
-        header_params['Accept'] = self.client.select_header_accept([
+
+        # User didn't specify the Accept header. Default it to json and yaml.
+        if not 'Accept' in header_params.keys():
```
roycaihw, line 224: you don't need `.keys()`, right?

Yashks1994: I find .keys() is the standard way, otherwise sometimes interpreter will throw some errors.

yliaog:
feel free to remove the hold once the case sensitivity is fixed

dynamic/client.py
```diff
@@ -219,8 +219,10 @@ def request(self, method, path, body=None, **params):
         header_params = params.get('header_params', {})
         form_params = []
         local_var_files = {}
-        # HTTP header `Accept`
-        header_params['Accept'] = self.client.select_header_accept([
+
+        # User didn't specify the Accept header. Default it to json and yaml.
+        if not 'Accept' in header_params:
```
roycaihw, line 224: my bad. Header names are not case sensitive, so it should be:
```
if not 'Accept' in header_params or not 'accept' in header_params:
```

Yashks1994: I run both by passing 'Accept' and 'accept', it ran fine.
```
{'Accept': 'application/json;as=PartialObjectMetadataList;v=v1;g=meta.k8s.io'}
PartialObjectMetadataList
meta.k8s.io/v1

{'accept': 'application/json;as=PartialObjectMetadataList;v=v1;g=meta.k8s.io'}
PartialObjectMetadataList
meta.k8s.io/v1 
```
Still I will make the necessary change.

roycaihw:
I think the next steps are:
1. find a solution to do case insensitive search for a key in dict. There are some options in https://stackoverflow.com/questions/3296499/case-insensitive-dictionary-search
2. add more test cases to make sure things like `Accept` `accept` `aCCEpt` etc work

dynamic/client.py
```diff
@@ -219,11 +219,16 @@ def request(self, method, path, body=None, **params):
...
-        # HTTP header `Accept`
-        header_params['Accept'] = self.client.select_header_accept([
-            'application/json',
-            'application/yaml',
-        ])
+
+        # Checking Accept header if True pass the user header else default it to json and yaml.
+        new_header_params = dict((key.lower(), value) for key, value in header_params.items())
+        if 'accept' in new_header_params:
+            header_params['Accept'] = new_header_params['accept']
```
roycaihw, line 226: if a user passes in an "accept" header, this will result in duplicated "Accept" and "accept" headers. You can simply do
```
if not 'accept' in new_header_params:
    header_params['Accept'] = self.client.select_header_accept([
        'application/json',
        'application/yaml',
    ])
```

Yashks1994: Done

Issue #239, opened 2021-05-26
[watch/iter_resp_lines] If the pod log contains blank lines, blank lines are ignored


FuZer:
https://github.com/kubernetes-client/python-base/blob/b87a5fe119b820e9196f39343b0f5a1546d7465b/watch/watch.py#L44-L58

it seems blank lines are ignored in pod logs.

when pod logs like below, this code ignore blank lines
```
poobar

poobar

poobar
```

```
yield "poobar"
yield "poobar"
yield "poobar"
```

but it have to be like below not ignore blank lines
```
yield "poobar"
yield ""
yield "poobar"
yield ""
yield "poobar"
```

PR #240, opened 2021-05-26
Fix watch ignore blank line in pod logs

FuZer:
This is relation to https://github.com/kubernetes-client/python-base/issues/239
In this PR, I fix watch ignore blank line in pod logs problem. 

with `enumerate` yield first line even if line is blank. when lines split by '\n' char.


yliaog:
watch is used for receiving events, not for logs. please clarify how the fix in watch could help with pod logs blank lines issue.

PR #241, opened 2021-05-28
Make duck-typing checking in `serialize_body` method more restrictive

piglei:
The`body` parameter in `DynamicClient`'s most methods currently accepts both dict type and `ResourceInstance` type. This was achieved by calling `serialize_body()` before mading any requests.

But in current implementation, the duck-typing style checking in `serialize_body` was kind of meaning-less---`ResourceInstance` type has overwritten it's `__getattr__` method, so basically it's instances can pass every `hasattr(...)` check.

```python
>>> inst = ResourceInstance(None, {'kind': 'Pod'})
>>> hasattr(inst, 'to_dict')
True
>>> hasattr(inst, 'to_dict_whatever_name')
True
```

This PR makes the checking in `serialize_body()` method more restrictive and also add some related unit tests.

### Release Note: 

```release-note
Type checking in `Client.serialize_body()` was made more restrictive and robust.
```

### DIFF
--- a/dynamic/client.py
+++ b/dynamic/client.py
@@ -98,7 +98,12 @@ class DynamicClient(object):
         return namespace
 
     def serialize_body(self, body):
-        if hasattr(body, 'to_dict'):
+        """Serialize body to raw dict so apiserver can handle it
+
+        :param body: kubernetes resource body, current support: Union[Dict, ResourceInstance]
+        """
+        # This should match any `ResourceInstance` instances
+        if callable(getattr(body, 'to_dict', None)):
             return body.to_dict()
         return body or {}
 
--- a/dynamic/test_client.py
+++ b/dynamic/test_client.py
@@ -20,6 +20,7 @@ from kubernetes.e2e_test import base
 from kubernetes.client import api_client
 
 from . import DynamicClient
+from .resource import ResourceInstance, ResourceField
 from .exceptions import ResourceNotFoundError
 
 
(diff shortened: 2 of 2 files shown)


piglei:
> /lgtm, needs rebase though

Rebased with upstream/master.

OWNERS
```diff
@@ -5,4 +5,5 @@ approvers:
   - roycaihw
 emeritus_approvers:
   - mbohlool
-
+reviewers:
+  - fabianvf
```
yliaog, line 9: there might be some problem with the rebase, this PR should not have this file

config/__init__.py
```diff
@@ -12,7 +12,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from os.path import exists, expanduser
```
yliaog, line 15: ditto, there might be some problem with the rebase, this PR should not have this file

OWNERS
piglei: It turns out that the rebase operation was correct, but somehow GitHub could not handle it properly, it still displays commits already exists in target branch.

Fixed by editing this PR to a different target branch and then switch it back. 

Ref: [git - GitHub pull request showing commits that are already in target branch - Stack Overflow](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch)

piglei:
MR refreshed.

yliaog:
please squash the commits

piglei:
> please squash the commits

Done

fabianvf:
@piglei just needs the release note and it should be good to go

piglei:
> @piglei just needs the release note and it should be good to go

Thanks for reminding me, the release note was now added.

PR #243, opened 2021-06-24
Add watch_stop to DynamicClient

hedinasr:
#### What type of PR is this?

/kind feature

This PR introduce a way to close a watching stream.

Fixes #

```release-note
NONE
```

### DIFF
--- a/dynamic/client.py
+++ b/dynamic/client.py
@@ -144,7 +144,7 @@ class DynamicClient(object):
 
         return self.request('patch', path, body=body, content_type=content_type, **kwargs)
 
-    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None):
+    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None):
         """
         Stream events for a resource from the Kubernetes API
 
@@ -156,6 +156,7 @@ class DynamicClient(object):
         :param resource_version: The version with which to filter results. Only events with
                                  a resource_version greater than this value will be returned
         :param timeout: The amount of time in seconds to wait before terminating the stream
+        :param watcher: The Watcher object that will be used to stream the resource
 
         :return: Event object with these keys:
                    'type': The type of event such as "ADDED", "DELETED", etc.
@@ -164,13 +165,17 @@ class DynamicClient(object):
 
         Example:
             client = DynamicClient(k8s_client)
+            watcher = watch.Watch()
             v1_pods = client.resources.get(api_version='v1', kind='Pod')
 
-            for e in v1_pods.watch(resource_version=0, namespace=default, timeout=5):
+            for e in v1_pods.watch(resource_version=0, namespace=default, timeout=5, watcher=watcher):
                 print(e['type'])
                 print(e['object'].metadata)
+                # If you want to gracefully stop the stream watcher
+                watcher.stop()
         """
-        watcher = watch.Watch()
+        if not watcher: watcher = watch.Watch()
+
         for event in watcher.stream(
             resource.get,
             namespace=namespace,


hedinasr:
Verify CLA

hedinasr:
> Thanks for your pull request. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.**
> 
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.
> 
> * If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address.  Check your existing CLA data and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * If you signed the CLA as a corporation, please sign in with your organization's credentials at https://identity.linuxfoundation.org/projects/cncf to be authorized.
> * If you have done the above and are still having issues with the CLA being reported as unsigned, please log a ticket with the Linux Foundation Helpdesk: https://support.linuxfoundation.org/
> * Should you encounter any issues with the Linux Foundation Helpdesk, send a message to the backup e-mail support address at: [login-issues@jira.linuxfoundation.org](mailto:login-issues@jira.linuxfoundation.org)

Check CLA

dynamic/client.py
```diff
@@ -169,20 +169,27 @@ def watch(self, resource, namespace=None, name=None, label_selector=None, field_
...
-            name=name,
-            field_selector=field_selector,
-            label_selector=label_selector,
-            resource_version=resource_version,
-            serialize=False,
-            timeout_seconds=timeout
-        ):
-            event['object'] = ResourceInstance(resource, event['object'])
-            yield event
+        self.watch_stop = watcher.stop
```
fabianvf, line 176: I think you can create multiple simultaneous watch invocations on the same client, but in this case doing that will mean that subsequent invocations before the initial watch is closed will overwrite this client-global variable. I think we'd need to find a way to scope the stop function the user can access to the specific watch that they need to stop.

dynamic/client.py
```diff
@@ -169,20 +169,27 @@ def watch(self, resource, namespace=None, name=None, label_selector=None, field_
...
+                name=name,
+                field_selector=field_selector,
+                label_selector=label_selector,
+                resource_version=resource_version,
+                serialize=False,
+                timeout_seconds=timeout
+            ):
+                event['object'] = ResourceInstance(resource, event['object'])
+                yield event
+        except KeyboardInterrupt as err:
```
fabianvf, line 190: is it typical to catch KeyboardInterrupts in a library like this? I would expect that to be handled by the caller. That being said, should we be stopping the watcher on any exception? Not sure if we gain anything from leaving it open if we're exiting this function anyway

fabianvf (changes requested):

dynamic/client.py
hedinasr: I totally agree with you on this one, this should be handled by the caller. Anyway, there is a `finally` clause in the `Watch.stream` function that cleans up the connection if anything bad happened, so my worries here are not relevant.

dynamic/client.py
hedinasr: What about adding a `watcher` optional parameter ? This way, the user can instantiate his own `Watch` object, and call the `stop` function from that object. An example:

```python
def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None):
    if not watcher: watcher = watch.Watch()
    # no changes here
```

```python
client = DynamicClient(k8s_client)
watcher = watch.Watch()
v1_pods = client.resources.get(api_version='v1', kind='Pod')

for e in v1_pods.watch(resource_version=0, namespace=default, timeout=5, watcher=watcher):
    print(e['type'])
    print(e['object'].metadata)
    watcher.stop()
```

What do you think ?

fabianvf: This approach seems sane to me, I can't really think of anything better

fabianvf (approved):
/lgtm

fabianvf (approved):

PR #244, opened 2021-06-27
Fix replication controller pods delete in tests

### LINKED ISSUE #242: ReplicationController delete propagation policy
When running the function test_replication_controller_apis from dynamic/test_client.py, the test deletes the replication controller, but not the underlying pods. I think this happens because by default the replication controller orphans the pods when deleted, and a solution could be to set the propagation policy to background or foreground.

If you want I can submit a PR here and then also in kubernetes-client/python, because there are replication controller tests in there that suffer from the same issue.

hedrox:
#### What type of PR is this?
/kind bug

Tests that create Replication Controller objects don't clean pods upon deletion

Fixes #242 

```release-note
NONE
```

```docs

```

### DIFF
--- a/dynamic/test_client.py
+++ b/dynamic/test_client.py
@@ -318,8 +318,10 @@ class TestDynamicClient(unittest.TestCase):
         self.assertEqual(name, resp.metadata.name)
         self.assertEqual(2, resp.spec.replicas)
 
-        resp = api.delete(
-            name=name, body={}, namespace='default')
+        api.delete(
+            name=name,
+            namespace='default',
+            propagation_policy='Background')
 
     def test_configmap_apis(self):
         client = DynamicClient(api_client.ApiClient(configuration=self.config))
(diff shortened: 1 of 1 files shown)


dynamic/test_client.py
```diff
@@ -318,8 +318,25 @@ def test_replication_controller_apis(self):
...
+                label_selector='name={}'.format(
+                    resp.spec.selector.name))
+            if not pods.items:
+                break
+            time.sleep(2)
+
+        self.assertEqual(
+            [],
+            pods.items,
+            'ReplicationController pods were not deleted in the 60 sec interval')
```
yliaog, line 339: 'Background' would do the pod deletion in the background, it does not guarantee that the pods are deleted in specific amount of time. 60s is arbitrary, the pods may, or may not be deleted during that interval. hence the test is flaky.

hedrox: Is it ok if I remove the check that the pods have been deleted? This would leave the test as it was, but now it would have the propagation policy fix. Any other solution that I have involves waiting forever either for the pods in case of the 'Background' policy, or for the ReplicationController in case of the 'Foreground' policy.

yliaog: have you verified manually that the pods are deleted eventually with 'Background'?

hedrox: Yes, I have. If you want I could change it to 'Foreground', but from my manual tests both the policies work.

yliaog: I think 'Background' is fine.

hedrox:
@roycaihw please tell me if there is anything I can do to help the review process. Thank you for your time.

PR #246, opened 2021-07-07
Fix load_config: expand ~

david0:
#### What type of PR is this?
/kind bug

This fixes loading config using the new `load_config` introduced with #221 from `~/.kube/config`.
`os.file.exists` does not perform some implicit expandsion of `~` and hence returns `False` for `os.path.exists('~/.kube/config')`

```release-note

```

```docs

```

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -45,7 +45,7 @@ except ImportError:
     pass
 
 EXPIRY_SKEW_PREVENTION_DELAY = datetime.timedelta(minutes=5)
-KUBE_CONFIG_DEFAULT_LOCATION = os.path.expanduser(os.environ.get('KUBECONFIG', '~/.kube/config'))
+KUBE_CONFIG_DEFAULT_LOCATION = os.environ.get('KUBECONFIG', '~/.kube/config')
 ENV_KUBECONFIG_PATH_SEPARATOR = ';' if platform.system() == 'Windows' else ':'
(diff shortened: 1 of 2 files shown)


config/kube_config.py
```diff
@@ -45,7 +45,7 @@
     pass
 
 EXPIRY_SKEW_PREVENTION_DELAY = datetime.timedelta(minutes=5)
-KUBE_CONFIG_DEFAULT_LOCATION = os.environ.get('KUBECONFIG', '~/.kube/config')
+KUBE_CONFIG_DEFAULT_LOCATION = os.path.expanduser(os.environ.get('KUBECONFIG', '~/.kube/config'))
```
yliaog, line 48: os.path.expanduser(path) is called at L708:
https://github.com/kubernetes-client/python-base/blob/dca0ca6df23098b63a347e12ceb0dca028a20572/config/kube_config.py#L708

is the call os.path.expanduser still needed?

david0: At least it would be a little bit of a breaking change. KubeConfigMerger and  load_kube_config et al are part of the API  would not take paths with `~` anymore. 

Maybe its clearer overall to do the expand_path in `load_config` to avoid doing twice in this case?

yliaog: sounds good to do the expand_path in  load_config

david0: done

PR #250, opened 2021-08-27
Refresh exec-based API credentials when they expire

emenendez:
#### What type of PR is this?
/kind bug

As described in kubernetes-client/python#741, some of the authentication schemes supported by Kubernetes require updating the client's credentials from time to time. The Kubernetes Python client currently does not support this, except for when using the `gcp` auth scheme. This is because the OpenAPI-generated client code does not generally expect credentials to change after the client is configured.

However, in OpenAPITools/openapi-generator#3594, the OpenAPI generator added a (undocumented) hook on the `Configuration` object which provides a method for the client credentials to be refreshed as needed. Now that this hook exists, the `load_kube_config()` function, used by the Kubernetes API to set up the `Configuration` object from the client's local k8s config, just needs to be updated to take advantage of this hook.

This patch does this for `exec`-based authentication, which should resolve kubernetes-client/python#741.

Also, as noted above, `load_kube_config()` already has a special-case monkeypatch to refresh GCP tokens. I presume this functionality was added before the OpenAPI generator added support for the refresh hook. This patch also refactors the GCP token refreshing code to use the new hook instead of the monkeypatch.

Tests are also updated.

This is a fix for kubernetes-client/python#741, but we shouldn't resolve that issue until this change is merged to `master` and the submodule ref is updated in kubernetes-client/python.

```release-note
Fixed kubernetes-client/python#741, an issue which prevented Kubernetes cluster api-tokens from exec-plugin auth providers from being refreshed after expiry.
```

N/A

```docs

```

### DIFF
--- a/config/kube_config.py
+++ b/config/kube_config.py
@@ -359,6 +359,8 @@ class KubeConfigLoader(object):
             self._refresh_gcp_token()
 
         self.token = "Bearer %s" % provider['config']['access-token']
+        if 'expiry' in provider['config']:
+            self.expiry = parse_rfc3339(provider['config']['expiry'])
         return self.token
 
     def _refresh_gcp_token(self):
@@ -483,8 +485,7 @@ class KubeConfigLoader(object):
             status = ExecProvider(self._user['exec']).run()
             if 'token' in status:
                 self.token = "Bearer %s" % status['token']
-                return True
-            if 'clientCertificateData' in status:
+            elif 'clientCertificateData' in status:
                 # https://kubernetes.io/docs/reference/access-authn-authz/authentication/#input-and-output-formats
                 # Plugin has provided certificates instead of a token.
                 if 'clientKeyData' not in status:
@@ -504,10 +505,13 @@ class KubeConfigLoader(object):
                     file_base_path=base_path,
                     base64_file_content=False,
                     temp_file_path=self._temp_file_path).as_file()
-                return True
-            logging.error('exec: missing token or clientCertificateData field '
-                          'in plugin output')
-            return None
+            else:
+                logging.error('exec: missing token or clientCertificateData '
+                              'field in plugin output')
+                return None
+            if 'expirationTimestamp' in status:
+                self.expiry = parse_rfc3339(status['expirationTimestamp'])
+            return True
         except Exception as e:
             logging.error(str(e))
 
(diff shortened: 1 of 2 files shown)


config/kube_config_test.py
```diff
@@ -956,17 +961,15 @@ def test_load_user_token(self):
...
         KubeConfigLoader(
             config_dict=self.TEST_KUBE_CONFIG,
             active_context="gcp",
             get_google_credentials=lambda: _raise_exception(
                 "SHOULD NOT BE CALLED")).load_and_set(fake_config)
         # Should now be populated with a gcp token fetcher.
-        self.assertIsNotNone(fake_config.get_api_key_with_prefix)
+        self.assertIsNotNone(fake_config.refresh_api_key_hook)
         self.assertEqual(TEST_HOST, fake_config.host)
-        # For backwards compatibility, authorization field should still be set.
```
roycaihw, line 969: nit: why do we remove this comment?

config/kube_config_test.py
```diff
@@ -1490,31 +1522,28 @@ def test__get_kube_config_loader_dict_no_persist(self):
...
     # These tests guard against changes to the upstream configuration class,
-    # since GCP authorization overrides get_api_key_with_prefix to refresh its
-    # token regularly.
+    # since GCP and Exec authorization use refresh_api_key_hook to refresh
+    # their tokens regularly.
 
-    def test_get_api_key_with_prefix_exists(self):
-        self.assertTrue(hasattr(Configuration, 'get_api_key_with_prefix'))
+    def test_refresh_api_key_hook_exists(self):
+        self.assertTrue(hasattr(Configuration(), 'refresh_api_key_hook'))
```
roycaihw, line 1529: Why do we use `Configuration()` instead of `Configuration`?

config/kube_config.py
```diff
@@ -560,25 +564,15 @@ def _load_cluster_info(self):
...
         if 'token' in self.__dict__:
-            # Note: this line runs for GCP auth tokens as well, but this entry
-            # will not be updated upon GCP token refresh.
             client_configuration.api_key['authorization'] = self.token
+
+            def _refresh_api_key(client_configuration):
+                if ('expiry' in self.__dict__ and _is_expired(self.expiry)):
+                    self._load_authentication()
+                    self._set_config(client_configuration)
+            client_configuration.refresh_api_key_hook = _refresh_api_key
```
roycaihw, line 575: Is the indentation correct? Do we want to install the hook only when `token` is set?

roycaihw (commented):
Thanks! LGTM in general.

LS80:
Thanks for the fix. Is it likely to be approved and merged?

roycaihw:
/lgtm
/approve

Sorry for the late reply. Thank you for the fix!

PR #251, opened 2021-09-03
fix: field extra_args recursive growth caused by Resource and Subreso…

jamesgetx:
#### What type of PR is this?
/kind bug

Due to the bug of the `to_dict` method in `Resource` and `Subresource`, field `extra_args` grow recursively when cache with `CacheDecoder`. The cache file fragment is shown below.

image

the pr also fix another problem which `Resource` lost valid attribute `singularName` and `shortNames` when `CacheDecoder` decode from cache file.


fabianvf (approved):
/lgtm

yliaog:
please squash the commits

jamesgetx:
> please squash the commits
Done

PR #252, opened 2021-09-16
fix: Ignore NotFoundError in the first call of LazyDiscoverer. __search

schneesu:
#### What type of PR is this?

/kind bug

According to fabianvf's suggestion, I try to catch the exception in the first call and safely ignore it

https://github.com/kubernetes-client/python/issues/1536#issuecomment-920265516

Fixes https://github.com/kubernetes-client/python/issues/1536

```release-note
Ignore NotFoundError in the first call of LazyDiscoverer. __search
```


schneesu:
Some mistake happened. Please ignore this PR, I will create new one

PR #253, opened 2021-09-16
fix: ignore NotFoundError in the first call of LazyDiscoverer. __search

schneesu:
#### What type of PR is this?

/kind bug

According to fabianvf's suggestion, I try to catch the exception in the first call and safely ignore it

https://github.com/kubernetes-client/python/issues/1536#issuecomment-920265516

Fixes https://github.com/kubernetes-client/python/issues/1536

```release-note
Ignore NotFoundError in the first call of LazyDiscoverer. __search
```

### DIFF
--- a/dynamic/discovery.py
+++ b/dynamic/discovery.py
@@ -237,7 +237,11 @@ class LazyDiscoverer(Discoverer):
         return self.parse_api_groups(request_resources=False, update=True)['apis'].keys()
 
     def search(self, **kwargs):
-        results = self.__search(self.__build_search(**kwargs), self.__resources, [])
+        # In first call, ignore ResourceNotFoundError and set default value for results
+        try:
+            results = self.__search(self.__build_search(**kwargs), self.__resources, [])
+        except ResourceNotFoundError:
+            results = []
         if not results:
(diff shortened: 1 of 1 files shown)


schneesu:
CLA signature registered

dynamic/discovery.py
```diff
@@ -237,9 +238,13 @@ def api_groups(self):
         return self.parse_api_groups(request_resources=False, update=True)['apis'].keys()
 
     def search(self, **kwargs):
+        # In first call, ignore NotFoundError when raised
+        self.__ignore_not_found_error = True
```
fabianvf, line 242: It may make more sense to just put line 243 into a `try... except` block, as multiple calls to this method could happen concurrently in which case they could wrestle over the `self.__ignore_not_found_error` variable

schneesu: Great suggestion, try-except is safer than use variable.

schneesu: PR refreshed.

schneesu:
Hello @yliaog , does this PR still need to be modified? It seems that the `lgtm` tag is required for robot auto-merge.

image

PR #254, opened 2021-09-16
add support for server side apply

abikouo:
#### What type of PR is this?
/kind feature

add support for server side apply feature for dynamic client

Fixes #

```release-note
NONE
```

```docs

```


abikouo:
@fabianvf

dynamic/client.py
```diff
@@ -136,15 +136,19 @@ def replace(self, resource, body=None, name=None, namespace=None, **kwargs):
         path = resource.path(name=name, namespace=namespace)
         return self.request('put', path, body=body, **kwargs)
 
-    def patch(self, resource, body=None, name=None, namespace=None, **kwargs):
+    def patch(self, resource, body=None, name=None, namespace=None, server_side=None, **kwargs):
```
fabianvf, line 139: thoughts on creating a separate apply method rather than reusing patch? Functionally this makes sense since it is just a specialized patch but I wonder if it would be more obvious to users to add it as a separate method (similar to kubectl apply vs patch)

abikouo: @fabianvf thanks for the feedback
new method ``apply`` created, please feel free to review again

dynamic/client.py
```diff
@@ -149,6 +149,23 @@ def patch(self, resource, body=None, name=None, namespace=None, **kwargs):
...
+    def apply(self, resource, body=None, name=None, namespace=None, server_side=None, force_conflicts=None, **kwargs):
+        body = self.serialize_body(body)
+        name = name or body.get('metadata', {}).get('name')
+        if not name:
+            raise ValueError("name is required to patch {}.{}".format(resource.group_version, resource.kind))
+        if resource.namespaced:
+            namespace = self.ensure_namespace(resource, namespace, body)
+
+        content_type = 'application/strategic-merge-patch+json'
+        if server_side is True:
```
fabianvf, line 161: Since this doesn't include an implementation of client-side apply, I think we can just remove the `server_side` argument and force the content_type to be `application/apply-patch+yaml` if this method is called

abikouo: updated

fabianvf (approved):
/lgtm

May need a release note though (that can be done in the PR text, doesn't require a commit or anything)

Issue #53, opened 2018-03-25
stream/ws_client.py doesn't close the connection if _preload_content is True


saurabh-chordiya:
Using kubernetes-4.0.0 python client and exec call as below:-

stream( api.connect_get_namespaced_pod_exec,                                                        
            name=pod_name, namespace=namespace, command=cmd,                                                         
            stderr=True, stdin=False, stdout=True, tty=False)

After this call it returns output as a string and not closing connection so If I use it in scale env then I see below error:-
filedescriptor out of range in select

So IMO ws_client.py should be fixed from below code snippet to close connection before return:-
        client.run_forever(timeout=_request_timeout)
        return WSResponse('%s' % ''.join(client.read_all()))

Please fix this if my observation is correct here.

saurabh-chordiya:
@mbohlool

bboe:
Here's a simple example that demonstrates the problem generically with select:

```python
#!/usr/bin/env python3
import select
import sys


def main():
    file_objects = []
    # Show that 1023 is okay.
    while not file_objects or file_objects[-1].fileno() < 1024:
        file_objects.append(open('/dev/null'))

    # FD 1023 works
    print(select.select((file_objects[-2],), (), ()))
    # FD 1024 does not
    print(select.select((file_objects[-1],), (), ()))


if __name__ == "__main__":
    sys.exit(main())
```

I verified that on both OS X and Linux the following exception occurs:

```
Traceback (most recent call last):
  File "./select_demo.py", line 19, in <module>
    sys.exit(main())
  File "./select_demo.py", line 15, in main
    print(select.select((file_objects[-1],), (), ()))
ValueError: filedescriptor out of range in select()
```

I'm going to work on a hopefully simple PR to change from select to poll as I just started running into this issue when using this library within a flask application that supports thousands of threads.

bboe:
Just to add to the above, the root issue is with whatever the file descriptor number is that [is being selected on](https://github.com/kubernetes-client/python-base/blob/1d5231cee6d0674b958ad429b60dda39363a4dc1/stream/ws_client.py#L174). In my case, this is due to external open files, but for you it could be due to the reason you've described. Either way, using select seems like a non-ideal way to go.

PaulFurtado:
@bboe it's not actually the file descriptor number being selected on, but the ulimit on the total count. The default ulimit for open files is 1024. File descriptor numbers are per-process, but ulimits account for all processes running as a given user.

[...]

PR #268, opened 2021-11-29
Use select.poll() for exec on linux/darwin

jsun-splunk:
#### What type of PR is this?

/kind bug

For an **exec** call, the current `WSClient` implementation uses `select.select()` to wait for file descriptor that are ready for read. The `select.select()` method uses the `select()` system call for this. As the `select()` system call has a limitation where by it will return the following error if the [file descriptor # is greater than 1024](https://man7.org/linux/man-pages/man2/select.2.html).

```
Reason: filedescriptor out of range in select()
```

This limitation happens at scale when are running a large number of **exec** calls from a single client process. Once the filedescriptor number goes pass 1024 all future **exec** calls will experience the above error.

By switching to [select.poll()](https://docs.python.org/3.8/library/select.html#polling-objects) instead, we can over come this limitation. The caveat here is that `select.poll()` is only supported on unix based OS, where as `select.select()` is also support on Windows. However, even if this is the case, it would be better to use the `select.poll()` where possible so we can overcome this limitation even if it is on unix based OS.

Similar problem was previously discussed in these issues. 
https://github.com/kubernetes-client/python-base/issues/53
https://github.com/kubernetes-client/python-base/issues/106

#### Special notes for your reviewer:

Another option here is to use `select.epoll()` instead of `select.poll()` for linux, however, I did not fully understand the difference between the two and did not felt confident enough to try epoll.

NONE

```release-note
Use select.poll() for exec on linux/darwin to improve scalability of WSClient
```

```docs

```

### DIFF
--- a/stream/ws_client.py
+++ b/stream/ws_client.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import sys
 
 from kubernetes.client.rest import ApiException, ApiValueError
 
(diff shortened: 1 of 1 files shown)


roycaihw:
/assign

Thanks for sending the fix! Please let us know when it's ready for review.

jsun-splunk:
Hi @roycaihw - thanks for looking into this. I'm just waiting on some internal process to sign off the CLA. The PR is actually ready to be reviewed.

jsun-splunk:
I signed it

jsun-splunk:
i signed it

roycaihw:
The CLA check still doesn't pass. Please make sure you've signed the CLA, and the email matches the one used in the Git commits.

jsun-splunk:
hi @roycaihw,

Now this is merged, do you know how we can update the submodule in the top leve client (https://github.com/kubernetes-client/python/tree/master/kubernetes)? I can't seem to find any instructions on how to release the submodules.

Thanks!

roycaihw:
Please follow [this instruction](https://github.com/kubernetes-client/python/blob/master/devel/submodules.md#update-submodule) to update the submodule and create a PR. Thanks!

PR #269, opened 2021-12-09
Async Request fix for Dynamic Client

venukarnati92:
/kind bug

What this PR does / why we need it?

async creation of custom resources (CR) using the Custom Resource Definition is failing with the following error

`AttributeError: 'ApplyResult' object has no attribute 'data'`

Fix for the issue:

For Async call meta_request(func) is failing because response object is different for async vs sync call.The following fix will make sure that response object is same for synchronous and asynchronous calls.

Which issue(s) this PR fixes:
https://github.com/kubernetes-client/python/issues/1626

Special notes for reviewer:

Usually for all async calls we get the response when we make the following call thread.get() but incase of dynamic client async calls I don't see that behavior because of the meta_request(func) is already parsing the response.

Does this PR introduce a user-facing change?
NONE

### DIFF
--- a/dynamic/client.py
+++ b/dynamic/client.py
@@ -267,20 +267,36 @@ class DynamicClient(object):
         # Authentication setting
         auth_settings = ['BearerToken']
 
-        return self.client.call_api(
-            path,
-            method.upper(),
-            path_params,
-            query_params,
-            header_params,
-            body=body,
-            post_params=form_params,
-            async_req=params.get('async_req'),
-            files=local_var_files,
-            auth_settings=auth_settings,
-            _preload_content=False,
-            _return_http_data_only=params.get('_return_http_data_only', True)
-        )
+        if params.get('async_req'):
+            return self.client.call_api(
+                path,
+                method.upper(),
+                path_params,
+                query_params,
+                header_params,
+                body=body,
+                post_params=form_params,
+                async_req=params.get('async_req'),
+                files=local_var_files,
+                auth_settings=auth_settings,
+                _preload_content=False,
+                _return_http_data_only=params.get('_return_http_data_only', True)
+            ).get()
+        else:
+            return self.client.call_api(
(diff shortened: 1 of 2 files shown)


dynamic/client.py
```diff
@@ -267,20 +267,36 @@ def request(self, method, path, body=None, **params):
...
+                query_params,
+                header_params,
+                body=body,
+                post_params=form_params,
+                async_req=params.get('async_req'),
+                files=local_var_files,
+                auth_settings=auth_settings,
+                _preload_content=False,
+                _return_http_data_only=params.get('_return_http_data_only', True)
+            ).get()
```
yliaog, line 284: the only difference is .get(), so better to assign self.client.call_api(...) to a var, and return it directly for sync call, and call .get() for async call

venukarnati92: Assigned value to a Variable.Please have a look at it

dynamic/client.py
```diff
@@ -281,6 +281,10 @@ def request(self, method, path, body=None, **params):
             _preload_content=False,
             _return_http_data_only=params.get('_return_http_data_only', True)
         )
+        if params.get('async_req'):
+            return api_response.get()
+        else:
+            return api_response
```
yliaog, line 287: thanks for the fix, could you add a  unittest to verify the fix?

venukarnati92: Sorry, for the delay.Took sometime to add my Linux Foundation ID to my company "Workday" corporate open source contributions list.

Added unittest for async.

venukarnati92:
@yliaog : Can you please review my PR and let me know if I need to make any changes.

yliaog:
thanks for the pr, it looks good.

now the https://github.com/kubernetes-client/python-base repo has been merged into https://github.com/kubernetes-client/python, could you please create a PR in that repo with the commits here?

PR #275, opened 2021-12-23
Run kubeconfig exec commands in the correct directory

aschleck:
#### What type of PR is this?

/kind bug

Kubeconfigs with exec specifications should invoke them so that referenced relative paths are relative to the config itself: https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#file-references . The current implementation in this library does not do so, so this PR fixes it.

I am not sure how to fix the config/kube_config_test.py tests because they patch `kubernetes.config.kube_config.ExecProvider` rather than the `ExecProvider` implementation in this PR, and for a similar reason I suspect the current `from kubernetes.config.exec_provider import ExecProvider` in `config/kube_config.py` will cause problems. But once this submodule is updated I think this should work fine, so not sure if there's anything I need to do.

I didn't find a filed bug about this.

```release-note
NONE
```

### DIFF
--- a/config/exec_provider.py
+++ b/config/exec_provider.py
@@ -31,7 +31,7 @@ class ExecProvider(object):
     * caching
     """
 
-    def __init__(self, exec_config):
+    def __init__(self, exec_config, cwd):
         """
         exec_config must be of type ConfigNode because we depend on
         safe_get(self, key) to correctly handle optional exec provider
@@ -53,6 +53,7 @@ class ExecProvider(object):
                 value = item['value']
                 additional_vars[name] = value
             self.env.update(additional_vars)
+        self.cwd = cwd
 
     def run(self, previous_response=None):
         kubernetes_exec_info = {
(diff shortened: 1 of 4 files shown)


aschleck:
I believe I've signed the CLA

yliaog:
kubernetes/config -> base/config (https://github.com/kubernetes-client/python/blob/master/kubernetes/config), kubernetes/config is a link to base/config, so kubernetes.config.kube_config.ExecProvider is the ExecProvider in this PR.

could you add tests to verify the fix works?

aschleck:
Sorry I am not very clear on the workflow with this repository. When making this PR I cloned python-base/ only, are you saying that I should I clone the python/ repository as a whole and then somehow make the PR by modifying the submodule?

When I looked into adding tests for kube_config.py, it seemed difficult because the existing ExecProvider tests don't actually test anything about the ExecProvider. Would the best thing be to see if I can add a side-effect to the run() invocation and maybe be able to verify the property is set correctly? Are there better ways? Obviously I already added tests for the ExecProvider itself.

yliaog:
this PR is fine, no need to create a new PR. when you try to run the tests locally, you may need to clone the python repo with the base submodule.

since kube_config_test is a unittest, i think it is fine to do what you suggested, i.e. add a side-effect to the run() invocation and verify the property is set correctly

aschleck:
Added a test to kube_config_test.py

